PR #15, opened 2015-04-03
added _optional_ attribute ID to threshold

mwalzer:
To be discussed, if optional should be mandatory and if we should integrate ID into CVParamType base


mwalzer:
might resolve https://github.com/qcML/qcML-development/issues/14

PR #21, opened 2016-04-28
Feature/metrics drafts

mwalzer:
based on M. Eisenacher's table from PSI-Meeting.
collect metrics, describe them


mwalzer:
we might consider breaking that list up into several. Maybe keeping a list of all with their names and links to .md files describing the single metric in more detail.

PR #47, opened 2018-10-24
JSON Schema

bittremieux:
I created a JSON schema specification to validate mzQC JSON files. Please check it out and let me know if you have any comments.

The JSON schema file is `mzqc_0_0_10.schema.json`. I also adapted Dave's most recent example file `20181005-1091_Pool_start_v0.2.qc.json` so that it's valid against this schema to have an example file.

There are a few remarks I had while creating the JSON schema which should be discussed:

- In the JSON examples we've only included a single `runQuality` so far. In contrast, the XML schema allowed multiple runQualities in a single `qcML` element. Either (1) we add a list of `run`/`setQualities` that can have one or multiple `run`/`setQuality` children, or (2) we allow only a single `run`/`setQuality` per mzQC file. I've used option 1 now, but it adds an extra unnecessary level of indirection if there's only a single `run`/`setQuality`, which will likely be the most prevalent case. Thoughts?

- JSON doesn't really have unique IDs. This informs a few changes and attention points.
    - I have changed the CV list from an actual list of CV elements where the ID is a property of each CV element to a parent element with each CV as property for its ID as key. This makes CV keys unique as properties of an element should be unique (although this is not enforced by all parsers).
    - Unfortunately I haven't been able to figure out how reference to the CVs in the cvParameter elements can be validated, i.e. that those CVs should exist. I don't think this will be possible using JSON schema, so our validator (TBD) should check for this explicitly.
    - Other than for the CVs, as they get referenced by cvParameters, I've omitted the previous ID XML attributes. I'm not sure whether we actually needed those as the name/accession of the qualityParameters can be used to identify the parameters, the IDs were mostly required by XML but not really used by us I think?

- Some small housekeeping changes: consistent element naming (no underscores, camel case), try to use proper JSON instead of XML-inspired JSON.

### DIFF
--- a/schema/v0_0_10/examples/20181005-1091_Pool_start_v0.2.qc.json
+++ b/schema/v0_0_10/examples/20181005-1091_Pool_start_v0.2.qc.json
@@ -1,274 +1,252 @@
 {
     "qcML": {
-        "_version": "0.1.0",
-	"__comment": "The above is a speculative schema version since we haven't updated in some time.",
-        "runQuality": {
-            "metaDataParameters": {
-                "InputFiles": [
-                    {
-                        "RawFile": {
-                            "_id": "rawfile001",
-                            "_location": "file://c:/Research/20171124-Lizex-Chia/1091_Pool_start.mzML",
-                            "FileFormat": {
-                                "_cvRef": "MS",
-                                "_accession": "MS:1000584",
-                                "_name": "mzML format"
-                            },
-                            "FileProperties": [
-                                {
-                                    "_cvRef": "MS",
-                                    "_accession": "MS:1000568",
-                                    "_name": "MD5",
-                                    "_value": "b583f6d2a91b4749d5a75885330f6e5d"
+        "version": "0.0.1",
+        "runQuality": [
+            {
+                "metadata": {
+                    "inputFiles": [
+                        {
+                            "rawFile": {
+                                "location": "file:///c:/Research/20171124-Lizex-Chia/1091_Pool_start.mzML",
+                                "fileFormat": {
+                                    "cvRef": "MS",
+                                    "accession": "MS:1000584",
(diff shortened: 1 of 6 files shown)


mwalzer:
CV checking would be a case for semantic validation then?

mwalzer:
I will update my example to this PR and may add some more comments.

mwalzer:
But validation would be handy, for both development and adding CI to our repository :recycle:

mwalzer:
@bittremieux Can you add an example validation, maybe in a readme.md? And which tools and commands to use?

bittremieux:
> CV checking would be a case for semantic validation then?

Yes indeed. We should make a list of all things that still have to be validated after the syntactic validation:

- Check that the name, accession (unit) references to CV elements in `qualityParameter`s correspond to the information in the CV.
- Check that all `cvRef`s link to valid `cv`s in the file.
- Check that `qualityParameter`s are unique within a `run`/`setQuality`.
- Check that filenames are unique within a `run`/`setQuality`.
- Check that multi-file metrics refer to existing filenames.
- Check that quantile metrics sum to 1.
- Check that the dimensions of multi-dimensional metrics are consistent.

I have migrated this to issue #51 as well to have a clear overview of all the checks we need to build into our validator.

bittremieux:
> @bittremieux Can you add an example validation, maybe in a readme.md? And which tools and commands to use?

Yes, I will restructure this pull request a bit and add some more information. Based on the call, probably a new folder for schema version 0.0.11? We've significantly deviated from the 0.0.10 XML schema, so that seems sensible.

bittremieux:
Also, as you mentioned, I removed the leading underscores because I wasn't sure what they meant. How do you feel about them @mwalzer? An underscore in Python or C/C++ and other languages usually means a hidden variable, which isn't the case here.

PR #48, opened 2018-10-25
added example for multivariate qc in json

mwalzer:


bittremieux:
It took me a little while, but by repeating `RawFile` elements for each of the three input files (line 8–31–54) you also get invalid JSON because there are duplicate `RawFile` property keys.

A similar problem would occur if you have multiple general `InputFile`s. I might just remove that property key in the JSON schema and have a list of `inputFile`s instead. That makes the schema easier at least. Although having a raw file is then less explicit.

bittremieux:
I adapted the example to validate against the JSON schema because I wanted to make sure the schema supported multi-file information, see 4b1b13a.

I've split the three files over three different `runQuality`s, followed by a `setQuality` for the combined statistic.

When we do it like this in a single file, the question is what we need to specify as `inputFiles` for the `setQuality`? I've duplicated the `inputFiles` from the three preceding `runQuality`s now, but that's not entirely correct because the input is actually the information in the `runQuality` derived from those input files.
If the `runQuality`s would have been stored in individual mzQC files then we could refer to those files instead, but that's not possible in this case.
So how should we do that here?

Side note: probably the `inputFiles` for the `runQualities` should include some identification files as well because some of the QC metrics are ID-based.

mwalzer:
We should move schema to v0_1_0 from @bittremieux s v0_0_11 after adding more examples. (Also to stay in sync with CV.)

PR #53, opened 2019-01-22
Add files via upload

dtabb73:
[20190122-qc-cv-with-metric-relationships-to-psi-ms.obo.txt](https://github.com/HUPO-PSI/mzQC/files/2781775/20190122-qc-cv-with-metric-relationships-to-psi-ms.obo.txt)

I have added links for each IDFree metric to corresponding terms in PSI-MS.  Was it appropriate to use the "relationship" term in this context?


mwalzer:
I think so. I'll close this PR and open a CV issue for you. Let's put the CV on the next telco's agenda!

Issue #63, opened 2019-04-18
[CV] description broken for "MS2 density per quantile"


julianu:
The description for the term QC:4000062 "MS2 density per quantile" is broken.
Could you please provide the correct description @dtabb73 ?

dtabb73:
Hi, Julian.

We have some number of peaks observed in each MS/MS scan (called the “MS2-Density” in QuaMeter IDFree).  We report the 25%ile (“Q1”), median (“Q2”), and 75%ile (“Q3”) for all these peak counts from tandem mass spectra in the RAW, at present.  These are integers, unless we have an even number of spectra and the middle two are different in value.

“Characterizes the distribution of peak counts among tandem mass spectra” would be an appropriate description.

Thanks,
Dave

julianu:
Thanks!

Issue #57, opened 2019-03-11
[CV] Issue with cv-term values and the use of cv in mzQC json


mwalzer:
e.g. in corresponding lists, defined: "lists with n elements each, n specified by the CV param value"
If cv value is n, then where do the lists go? This contradicts, how we use e.g. id: QC:4000050
name: XIC-WideFrac

julianu:
Right, using the actual "cv-value" to  set the length of a tuple was my idea.
I was thinking about having a "cv-value" further describing the tuple and just have another "qc-metric-value" for the actual metric value. Whether the "cv-value" would be mandatory would be dependent on the actual CV, as would be the dimension (specifying a vector, matrix, table...).
Alternatively, the length could be given implicit by e.g. the vector length. But having it set explicitly adds another layer of control and correctness.

bittremieux:
Yes, I was thinking about this while working on the semantic validation as well. If you have an `n-tuple` metric the semantic validation could potentially check whether all `n` values are present in the mzQC file, but right now the number of values isn't specified yet.

julianu:
Decided in the PSI meeting:
We drop this redundant information and define the length of arrays/tables/matrices implicitly by the given json data.

Issue #54, opened 2019-02-21
[CV] PSI-MS relations to our entries (see #53)


mwalzer:
@dtabb73  added PSI-MS relations to the ID free metrics, e.g. `relationship: has_relation MS:1000041 ! charge state`.
[attached file from #53 ](https://github.com/HUPO-PSI/mzQC/files/2888627/20190122-qc-cv-with-metric-relationships-to-psi-ms.obo.txt)

julianu:
Added in the next OBO

PR #65, opened 2019-04-18
Updated some issues of the PSI meeting

julianu:
- added PSI MS relations (#54)
- removed length information from value (#57)
- added description from (#63)


bittremieux:
As for the question in point 3, I think the size of tables and matrices should similarly be defined implicitly.

Do we need to have the unit ontology obo in this repository as well? We're not maintaining it, so it'll get out of date. Why not rather keep it at its original location and refer to it there?

dtabb73:
Hi, folks.  As I understood it, we should have a CV 0.1.1 for me to use in updating my QuaMeter IDFree example.  Is there a reason this pull request hasn't been integrated?

mwalzer:
@dtabb73 : raised at least one point we should consider (regarding the uo.obo) and the other one I try to make sense of. @bittremieux which question in point 3 are you referring to? In fact, which point 3 - #63?

mwalzer:
@julianu : I agree with @bittremieux  on the uo.obo matter. There would be two solutions for a link if you need it in the MzQC structure (and I also see that it would be useful to have that one available, too, when you check out a local copy of MzQC):
* [submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules)
* [subtree](https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt)

mwalzer:
@julianu was [this](https://github.com/bio-ontology-research-group/unit-ontology) the origin of the `uo.obo`?

bittremieux:
@mwalzer The short question in point 3 in the readme.

mwalzer:
PR moved here: #67 
Further discussion there. You can also commit to that branch.

PR #67, opened 2019-05-28
Clone of: Updated some issues of the PSI meeting #65

mwalzer:
This is a clone of #65 in a communal branch of HUPO0-PSI/mzQC where all members can contribute
>
>- added PSI MS relations (#54)
>
>- removed length information from value (#57)
>
>- added description from (#63)


bittremieux:
No longer relevant; closing.

PR #69, opened 2019-07-18
MzQC pylib

mwalzer:
as preparations to push towards release 1.0
A python lib with which we can:
* start building MzQC objects and write then as json MzQC
* read MzQC files into object representation 
* make an assessment of syntactical validity 
* test and expand the set of semantic rules for our standard format release 1.0

needs closer alignment with schema element names (camel vs. snake case )


mwalzer:
This will be reopened from mzQC/mzqc-pylib

Issue #64, opened 2019-04-18
[CV] Why is ambient humidity an n-tuple?


julianu:
The term QC:4000068 "Ambient humidity" is currently an n-tuple.
Is there any reason for this? If so, the description (and name) should reflect this.

bittremieux:
It's an n-tuple because first I had this abominable value consisting of a (mean, median, sd, q1, q3, min, max) tuple. I think that by now we all agree that this isn't the way to encode this kind of information, rather each of these values should be an individual term. So this should be changed to a single value.

julianu:
Ok, thanks for the clarification.

PR #71, opened 2019-07-31
a first go at the OBO...

cbielow:
changes to be discussed:

I have made some changes to the OBO structure, in particular:
 - added `sample preparation metric` as child to QC metric
 - added `MS metric` as layer between `QC metric` and `MS1 metric` (same for MS2)
 - started changing some IDs:
    - the idea is that the IDs somewhat reflect the tree/graph structure, such that a knowledgable user ca n infer that `QC:4001????` must be a some sort of RT metric... not sure of this is a) allowed b) agreeable
 - the `ambient humidity` had a TODO (why n-tuple #64 ). I just came up with something completely different (corresponding lists), hoping that this would make sense. To be discussed I guess :)
 - some minor typos and low/uppercase fixes to make everything a bit more uniform
 - I removed (out of ignorance) the qc-cv-legacy-0.0.8.obo. Why was that present? It seems to me it should not carry over... Can someone enlighten me :)

I'll open up another issue with related questions...
PTXQC metrics are next on my list. Coming soon...

note: my `Protege` editor reorders some lines automagically. I hope thats ok...

### DIFF
--- a/cv/v0_1_0/qc-cv.obo
+++ b/cv/v0_1_0/qc-cv.obo
@@ -146,15 +146,37 @@ def: "A metric based on a chromatogram" [PSI:QC]
 is_a: QC:4000001 ! QC metric
 
 [Term]
+id: QC:4000023
+name: MS1 metric
+def: "a metric based on MS1 events" [PSI:QC]
+is_a: QC:4000028 ! QC metric
+
+[Term]
+id: QC:4000024
+name: MS2 metric
+def: "a metric based on MS2 events" [PSI:QC]
+is_a: QC:4000028 ! QC metric
+
+[Term]
 id: QC:4000025
 name: ion source metric
 def: "A metric related to events in the ion source" [PSI:QC]
-is_a: QC:4002000 ! MS metric
+is_a: QC:4000028 ! MS metric
 
 [Term]
 id: QC:4000026
 name: environment metric
 def: "A metric related to measurements of the environment/laboratory/room." [PSI:QC]
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:4000027
+name: sample preparation metric
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:4000028
+name: MS metric
 is_a: QC:4000001 ! QC metric
 
 [Term]
@@ -214,7 +236,7 @@ xref: value-type:xsd\:int "The allowed value-type for this CV term."
 is_a: QC:4000004 ! n-tuple
 is_a: QC:4000010 ! ID free
 is_a: QC:4000021 ! retention time metric
-is_a: QC:4002100 ! MS1 metric
+is_a: QC:4000023 ! MS1 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 
 [Term]
@@ -225,7 +247,7 @@ xref: value-type:xsd\:int "The allowed value-type for this CV term."
 is_a: QC:4000004 ! n-tuple
 is_a: QC:4000010 ! ID free
 is_a: QC:4000021 ! retention time metric
-is_a: QC:4002200 ! MS2 metric
+is_a: QC:4000024 ! MS2 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 
 [Term]
@@ -235,7 +257,7 @@ def: "The log ratio for the second to n-th quantile of TIC changes over first qu
 is_a: QC:4000004 ! n-tuple
 is_a: QC:4000010 ! ID free
 is_a: QC:4000022 ! chromatogram metric
-is_a: QC:4002100 ! MS1 metric
+is_a: QC:4000023 ! MS1 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 
 [Term]
@@ -245,7 +267,7 @@ def: "The log ratio for the second to n-th quantile of TIC over first quantile o
 is_a: QC:4000004 ! n-tuple
 is_a: QC:4000010 ! ID free
 is_a: QC:4000022 ! chromatogram metric
-is_a: QC:4002100 ! MS1 metric
+is_a: QC:4000023 ! MS1 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 
 [Term]
@@ -254,7 +276,7 @@ name: MS1 count
 def: "The number of MS1 events in the run." [PSI:QC]
 is_a: QC:4000003 ! single value
 is_a: QC:4000010 ! ID free
-is_a: QC:4002100 ! MS1 metric
+is_a: QC:4000023 ! MS1 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 
 [Term]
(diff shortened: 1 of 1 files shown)


cbielow:
image

bittremieux:
There is a simple resolution for issue #64. @julianu should just incorporate that in the next version of the CV.

Your list-of-lists solution as it is described at the moment doesn't seem ideal because ambient humidity will typically come from an external sensor, so there's no direct link to RT. At least this would be the case for my iMonDB software, where I report ambient humidity. I also only retain a single (average) ambient humidity value for each run to condense the amount of data.

mwalzer:
>>> removed (out of ignorance) the qc-cv-legacy-0.0.8.obo
the file or the include statement?

cbielow:
t

> 
> 
> > removed (out of ignorance) the qc-cv-legacy-0.0.8.obo
> 
> the file or the include statement?

the include
https://github.com/HUPO-PSI/mzQC/pull/71/files#diff-77d9fee4d2a21e086c38ee88e00ad9f4L7

mwalzer:
> a bit more uniform

I think we should define our style guide. #74

mwalzer:
> I just came up with something completely different (corresponding lists)
We were aware that there will be ambiguities, and I think this is a good place to discuss and document (I won't use a word as strong as guidelines here) our design intentions. It was mentioned in a previous call that if we have all different designs from different metric producers we take a cut in the added value of a common format. Having a well-discussed (specification document supplement?) rationale for our design decisions might help bring people to a common base.

mwalzer:
Further discussion point may be the separation between `chromatogram metric` and `retention time metric`

julianu:
> > removed (out of ignorance) the qc-cv-legacy-0.0.8.obo
> 
> only the import statement or the file also? I agree the legacy import shouldn't be imported, however the file itself should be available.

To keep a backlink to this file, the file was linked. We should keep the link there, if we need some kind of "backwards compatibility" with the old version. But if especially @mwalzer is happy to delete it, I would be perfectly fine to delete it as well.

> started changing some IDs: 
This might not be such a good idea. Usually, in the other PSI OBOs, the IDs are simply increasing numbers. To somewhat distinguish our CV from the legacy qcML-CVs, I put the 4 in the beginning. But I don't know whether we should use the IDs to categorize items. Maybe @Ozzard has any suggestions here?

mwalzer:
>  @mwalzer is happy to delete it,

Remove the mention of the legacy obo in our master obo? Fine by me. The legacy obo file itself we can ban to the `legacy` folder. We could see to avoid uid clashes, but that's sugar on top, don't bother if this complicates things

julianu:
> Remove the mention of the legacy obo in our master obo? Fine by me. The legacy obo file itself we can ban to the `legacy` folder. We could see to avoid uid clashes, but that's sugar on top, don't bother if this complicates things

Ok, then we can keep this out of the new overhauled OBO file.

Only the issue with increasing IDs needs to be discussed (in the upcoming call?). All other changes are ok for me.

Ozzard:
> > started changing some IDs:

> This might not be such a good idea. Usually, in the other PSI OBOs, the IDs are simply increasing numbers. To somewhat distinguish our CV from the legacy qcML-CVs, I put the 4 in the beginning. But I don't know whether we should use the IDs to categorize items. Maybe @Ozzard has any suggestions here?

I'm a great fan of allocating an ID for a concept exactly once and never, ever changing it; and treating an ID as having no meaning.  It's an opaque bundle of bits.

There are several reasons for this:
* Over time, the ID becomes embedded in bits of software, configuration files, tutorials, databases, and cross-links from other information.  Change the ID and all dependents are out of date.  Don't ever assume that just because you've trawled through all the information you can, you're in the clear.  See, for example, https://github.com/PaulBrack/Yamato/blob/master/SwaMe/FileMaker.cs .
* As you release new versions of controlled vocabularies, you tend to add new terms and retire old ones.  Even if your original terms fitted into whatever gaps you left in your numbering scheme, there's no guarantee you won't breach them with subsequent additions.  Almost the canonical example of this is version 2 of the Read Clinical Classification ("the Read Codes"), which used 4- and 5-character hierarchical codes with upper case, lower case and digits (so 62 options) available at each level of the hierarchy.  Read then promptly tried to code things like "job" (starting with "Prime Minister" and working down) and "interpreter required for <language>", and blew the namespace.  Read version 3, and the mapping to SNOMED, simply uses random term codes, with a side data structure to represent hierarchy, and is far better for that change.

Happy to discuss, but my working approach would be to attach no meaning to IDs and to distribute a separate file of groupings/hierarchy and another of equivalences if required.  Easier extension, easier versioning.

Ozzard:
> Only the issue with increasing IDs needs to be discussed (in the upcoming call?).

I'm not part of the WG, but would appreciate having some input into this decision if you don't have a librarian / taxonomist* / ontologist in the WG.

\* (Someone who maintains hierarchies of terms, not someone who stuffs animals - though the slight smell of preservation is common to the two)

julianu:
Thanks for the input, @Ozzard 

We will drop the categorisation-attempt of the IDs and just use them in increasing order.
I will change this in the next few days and accept the PR.

cbielow:
> 
> 
> > > started changing some IDs:
> 
> > This might not be such a good idea. Usually, in the other PSI OBOs, the IDs are simply increasing numbers. To somewhat distinguish our CV from the legacy qcML-CVs, I put the 4 in the beginning. But I don't know whether we should use the IDs to categorize items. Maybe @Ozzard has any suggestions here?
> 
> I'm a great fan of allocating an ID for a concept exactly once and never, ever changing it; and treating an ID as having no meaning. It's an opaque bundle of bits.
> 
> There are several reasons for this:
> 
>     * Over time, the ID becomes embedded in bits of software, configuration files, tutorials, databases, and cross-links from other information.  Change the ID and all dependents are out of date.  Don't ever assume that just because you've trawled through all the information you can, you're in the clear.  See, for example, https://github.com/PaulBrack/Yamato/blob/master/SwaMe/FileMaker.cs .

Point well taken, but the mzQC standard is not released yet, thus no one should be using the current IDs (or trust they are stable). With the first release, we should not ever touch them, of course.

> 
>     * As you release new versions of controlled vocabularies, you tend to add new terms and retire old ones.  Even if your original terms fitted into whatever gaps you left in your numbering scheme, there's no guarantee you won't breach them with subsequent additions.  Almost the canonical example of this is version 2 of the Read Clinical Classification ("the Read Codes"), which used 4- and 5-character hierarchical codes with upper case, lower case and digits (so 62 options) available at each level of the hierarchy.  Read then promptly tried to code things like "job" (starting with "Prime Minister" and working down) and "interpreter required for <language>", and blew the namespace.  Read version 3, and the mapping to SNOMED, simply uses random term codes, with a side data structure to represent hierarchy, and is far better for that change.
> 

Very interesting! Will keep that in mind.

> 
> Happy to discuss, but my working approach would be to attach no meaning to IDs and to distribute a separate file of groupings/hierarchy and another of equivalences if required. Easier extension, easier versioning.

Fair point. Thanks for the input. :)

PR #78, opened 2019-08-16
Mzqc pylib

mwalzer:
continuation of #70


bittremieux:
Are you including the Python library somewhere else or... ?

PR #82, opened 2019-09-10
flatten OBO versioning

cbielow:
get rid of extra directories with versioning.
GIT should take care of this just fine.
Increases usability (e.g. opening THE obo always gives you the latest file -- otherwise you might miss that there is a new folder) and prevents platform incompatibilities (symlink does not work on non-Linux OS').


cbielow:
ooopps.. Git is tracking symlinks using file mode 120000.
fixing...

cbielow:
ready for merge?!

julianu:
The only reason that let's me hesitate to merge this is, that it for some reason was started to "version" the files in the folder-wise-way.
Any views of the rest, or maybe let's put it on the agenda for the next call.

bittremieux:
This is a necessary change imo. All the different versions as separate files is unwieldy and confusing for people checking out the repository. Like Chris said, published CV versions can be designed with a git tag. This will also make it more clear for users that the CV reference in their mzQC files should be a specific commit, not just `master`.

The same directory flattening should actually also happen for the different mzQC schema folders we have right now.

Issue #97, opened 2020-02-06
[CV request] SmaMe metrics


julianu:
Requested Metrics by e-mail:

# SWATH divided metrics

### Name: Target m/z

Definition: Part of SwaMe SWATH division. The isolation window target m/z is read straight from the mzML file and reported for every unique isolation window.

How to interpret the metric value: This serves mainly as information to orientate the user.

Proposed value type:
- [ ] single value
- [x] n-tuple
- [ ] corresponding lists
- [ ] table
- [ ] matrix

Identifications needed (if appropriate):
- [ ] ID based
- [x] ID free

### Name: Scans per swath

Definition: Part of SwaMe SWATH division. The number of scans collected during the run reported for all unique isolation windows.

How to interpret the metric value: This is mainly informative and serves as a nice confirmation that mzML contains all the necessary information.

Proposed value type:
- [ ] single value
- [x] n-tuple
- [ ] corresponding lists
- [ ] table
- [ ] matrix

Identifications needed (if appropriate):
- [ ] ID based
- [x] ID free

### Name: Average m/z range of SWATH

Definition: Part of SwaMe SWATH division. The average range in m/z of each unique isolation window.

How to interpret the metric value: This is the range set by the user in the isolation scheme. The purpose is to provide context for other metrics.

Proposed value type:
- [ ] single value
- [x] n-tuple
- [ ] corresponding lists
- [ ] table
- [ ] matrix

Identifications needed (if appropriate):
- [ ] ID based
- [x] ID free

### Name: Swath proportion of total TIC

Definition: Part of SwaMe SWATH division. The TIC for all occurrences of the relevant isolation window in the run is summed and divided by the total TIC for all MS2 scans in the run.

How to interpret the metric value: A larger value indicates that swaths with that particular target m/z contributed more to the total TIC.

Proposed value type:
- [ ] single value
- [x] n-tuple
- [ ] corresponding lists
- [ ] table
- [ ] matrix

Identifications needed (if appropriate):
- [ ] ID based
- [x] ID free

### Name: Average SWATH density

Definition: Part of SwaMe SWATH division. For each unqiue isolation window this value reports the number of ions detected on average throughout the run.

How to interpret the metric value: A higher value indicates that a higher number of ions were detected for SWATHs of this specific isolation window on average.

Proposed value type:
- [ ] single value
- [x] n-tuple
- [ ] corresponding lists
- [ ] table
- [ ] matrix

Identifications needed (if appropriate):
[...]

PR #98, opened 2020-03-05
Added CV entries from issue #97

julianu:
Added the SwaMe metrics. Please comment and review.


bittremieux:
@julianu Is more information needed to add these metrics or are they ready?
@dtabb73 Is the latter even realistic as Marina has graduated?

I.e., can this be merged as is, or should we just close the PR and related issue for now without merging?

julianu:
It should be ok, if the metrics for SwaMe did not change (significantly) in the last few months.

Issue #101, opened 2020-06-01
Sampling rate CV entry


mwalzer:
Name: Sampling rates

Definition: The sampling rates of identified peptides and their respective frequency (accumulated by the sampling rate, i.e. how often peptides were sampled n times and how many peptides were resampled at that rate. Optionally, which peptides were sampled at that rate, space separated)

Comment: The sampling rate may give insights to peptide separation, dynamic exclusion settings and gradient efficiency.

Proposed value type:
- [ ] table
{'Sample rate':[...], 'count':[...], *'peptides':[...]}
*optional

Identifications needed (if appropriate):
- [ ] ID based

Issue #91, opened 2019-09-26
Protein coverage CV term


mwalzer:
Name:
Protein coverage

Definition:
The coverage of protein sequences from the peptide sequences identified. The table records the coverage itself (in percent), the protein accession (SHOULD correspond to the accession used in fasta document used for Sequence DB based identification), Length of the protein(, and optionally if it is 'target' or 'decoy' entry.)

Comment: The protein coverage can provide insight into the sensitivity of the instrument/identification/protocols used.

Proposed value type:
- [ ] table
{'coverage':[...], 'accession':[...], 'length':[...], *'TD':[...]}
*optional

mwalzer:
![examplenewplot](https://user-images.githubusercontent.com/5859205/65682253-7b2d0980-e052-11e9-8b9d-7f5680bdefbd.png)

julianu:
Please add comments (i.e. descriptions on how to interpret the values) for the metrics.

As we also have "table" as type, this would be a better suggestion in my opinion.

julianu:
@mwalzer please add comment (explanation for interpretation)

mwalzer:
Indeed, and I refined the definition a little.

Name:
Protein coverage

Definition:
The coverage of protein sequences from the peptide sequences identified. The table records the coverage itself (percentage), the protein accession (SHOULD correspond to the accession used in the document used for Sequence DB based identification), Length of the protein(, and optionally if it is a decoy entry.)

Comment:
Higher is better. Gives insight into how much of the target search space is covered. Low coverage can be caused by the absence of proteins, because of the used acquisition methods insensitivity to certain peptide species or shortcomings of the identification method.

Proposed value type:
- [ ] single value
- [ ] n-tuple
- [x] table
- [ ] corresponding lists
- [ ] matrix
    
{'coverage':[...], 'accession':[...], 'length':[...], *'TD':[...]}
    *optional

mwalzer:
@julianu Where would we put something like `relationship: has_order MS:1002108 ! higher score better`? As text in the comments is suboptimal. And how do we designate which column is primary/affected by `has_order`? And units.

mwalzer:
@julianu Maybe we can use this one as an example during PSI2020 to formalise the cv term value situation.

Issue #90, opened 2019-09-26
New batch of CV term requests, category: Identifications (distributions)


mwalzer:
Name:
Precursor errors (abs) Q1, Q2, Q3

Definition:
From the distribution of Precursor errors (abs), the quartiles Q1, Q2, Q3 value

Comment: The absolute precursor error distribution can aid the interpretation of overall identification success.

Proposed value type:
- [ ] n-tuple
{STATO:0000167, STATO:0000574, STATO:0000170}

---
Name:
Precursor errors (abs) mean

Comment: The absolute precursor error distribution can aid the interpretation of overall identification success.

Definition:
From the distribution of Precursor errors (abs), the mean

Proposed value type:
- [ ] single value
{STATO:0000401}

Synonym: MS1-5B

---
Name:
Precursor errors (abs) sigma

Definition:
From the distribution of Precursor errors (abs), the sigma value

Comment: The absolute precursor error distribution can aid the interpretation of overall identification success.

Proposed value type:
- [ ] single value
{STATO:0000237}

---
Name:
Precursor errors (abs) outliers (<Q1-1.5*IQR, >Q3+1.5*IQR)

Definition:
From the distribution of Precursor errors (abs), the list of outliers below of Q1-1.5IQR and above Q3+1.5IQR

Comment: The absolute precursor error distribution can aid the interpretation of overall identification success.

Proposed value type:
- [ ] n-tuple
{'NCIT:C75330':[...], 'NCIT:C75329':[...]}
{'below':[...], 'above':[...]}

---
Name:
Precursor errors (ppm) Q1, Q2, Q3

Comment: The precursor error in ppm distribution can aid the interpretation of overall identification success.

Definition:
From the distribution of Precursor errors (ppm), the quartiles Q1, Q2, Q3 value

Proposed value type:
- [ ] n-tuple
{STATO:0000167, STATO:0000574, STATO:0000170}

---
Name:
Precursor errors (ppm) mean

Definition:
From the distribution of Precursor errors (ppm), the mean 

Comment: The precursor error in ppm distribution can aid the interpretation of overall identification success.

Proposed value type:
- [ ] single value
{STATO:0000401}

---
Name:
Precursor errors (ppm) sigma

Definition:
From the distribution of Precursor errors (ppm), the sigma value

Comment: The precursor error in ppm distribution can aid the interpretation of overall identification success.

Proposed value type:
- [ ] single value
{STATO:0000237}

---
Name:
[...]

Issue #89, opened 2019-09-26
New batch of CV term requests, category: Identifications


mwalzer:
Name:
Charge 1+ to charge 2+ peptides

Definition:
Ratio of 1+ peptide count  over 2+ peptide count  in identified spectra

Proposed value type:
- [ ] single value
---
Name:
Charge 3+ to charge 2+ peptides

Definition:
Ratio of 3+ peptide count over 2+ peptide count in identified spectra

Proposed value type:
- [ ] single value
---
Name:
Charge 4+ to charge 2+ peptides

Definition:
Ratio of 4+ peptide count over 2+ peptide count in identified spectra

Proposed value type:
- [ ] single value
---
Name:
Mean charge in identified spectra

Definition:
Mean charge in identified spectra

Proposed value type:
- [ ] single value
---
Name:
Median charge in identified spectra

Definition:
Median charge in identified spectra

Proposed value type:
- [ ] single value
---
Name:
Charge 1+ to charge 2+ spectra

Definition:
Ratio of 1+ peptide count  over 2+ peptide count  in all MS2

Proposed value type:
- [ ] single value
---
Name:
Charge 3+ to charge 2+ spectra

Definition:
Ratio of 3+ peptide count over 2+ peptide count  in all MS2

Proposed value type:
- [ ] single value
---
Name:
Charge 4+ to charge 2+ spectra

Definition:
Ratio of 4+ peptide count over 2+ peptide count  in all MS2

Proposed value type:
- [ ] single value
---
Name:
Mean charge in all MS2

Definition:
Mean charge in all MS2

Proposed value type:
- [ ] single value
---
Name:
Median charge in all MS2

Definition:
Median charge in all MS2

Proposed value type:
- [ ] single value
---
Name:
Number of missed cleavages

Definition:
Number of missed cleavages

Proposed value type:
- [ ] single value
---
Name:
Total number of protein evidences

Definition:
Total number of protein evidences

Proposed value type:
- [ ] single value
---
Name:
Number of identified proteins

Definition:
Number of identified proteins

Proposed value type:
- [ ] single value
---
Name:
Total number of PSM

Definition:
Total number of PSM

Proposed value type:
- [ ] single value
---
Name:
Number of identified peptides

Definition:
Number of identified peptides

Proposed value type:
- [ ] single value
---
Name:
Number of identified spectra

Definition:
Number of  identified spectra

Proposed value type:
- [ ] single value
---
Name:
ID ratio

Definition:
[...]

PR #102, opened 2020-06-16
added cv term in bulk

mwalzer:
from #85, #86, #87, #88, #89, #90, #91, #101
* second pair of eyes for the metric types needed

### DIFF
--- a/cv/CV-guide.md
+++ b/cv/CV-guide.md
@@ -19,7 +19,7 @@ to define your metric object which would look like
 {
     "accession": "QC:4000111",
     "name": "Number of MS1 spectra",
-    "value": 7130,
+    "value": 7130
 }
 ```
 where the value would then be implicitly interpreted as of type `count`.
@@ -42,7 +42,7 @@ Then we can assign the tuples as an n-array and their units implicitly assumed t
 {
     "accession": "QC:4000113",
     "name": "MZ acquisition range",
-    "value": [400,1200],
+    "value": [400,1200]
 }
 ```
 
@@ -90,7 +90,7 @@ def: "The column contains peptide sequences." [PSI:QC]
 is_a: QC:4000082 ! Defined value type
 has_units: NCIT:C45253 ! String
 ```
-Such, a metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLVTHYDR") and three peptides sampled 3 times ("IQACSGNDLQQWAR","AWQQLDNGSCAQIR","MATSGMMTELGK") will look like this.
+A metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLVTHYDR") and three peptides sampled 3 times ("IQACSGNDLQQWAR","AWQQLDNGSCAQIR","MATSGMMTELGK") will look like this.
 ```
 {
     "accession": "QC:4000113",
@@ -98,23 +98,23 @@ Such, a metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLV
     "value": {
 	    "Sample rates": [5,3],
 	    "Counts": [2,3],
-	    "Peptide sequences": ["DYHTVLGAR AGLVTHYDR","IQACSGNDLQQWAR AWQQLDNGSCAQIR MATSGMMTELGK"],
+	    "Peptide sequences": ["DYHTVLGAR AGLVTHYDR","IQACSGNDLQQWAR AWQQLDNGSCAQIR MATSGMMTELGK"]
     },
 }
 ```
 
 
-# Define a new metric term for a
-If a desired metric is not yet defined, you will need a cv term definition before you can write a metric object. 
+# Define a new metric term for ...
+If a desired metric is not yet defined, you will need a cv term definition before you can write a metric object. First to decide is what type of value your new metric will have.
 
 ## single value, n-tuple, matrix:
 This is straight forward for these types. 
 1. Make sure, the term is not defined yet in the QC ontology
 2. Define a reasonable descriptive name while keeping concise
 3. Describe the term content, context (relations), and usage 
-4. Define its value type 
+4. Define its value type and unit 
 
-Say we want to define a metric for the fastest frequency for MS/MS acquisition. We look up the QC ontology and for the purpose of this document pretend, no such term exists. We coin the name `Maximal MS2 frequency` and define it as `"The fastest frequency for MS/MS collection in any minute over the complete run"`.  This will clearly be a single value type, we do not need any spectrum identifications to assign its value, but it will be for the MS2 tandem spectra of one particular run.  And it will clearly be of type frequency, in particular the sampling frequency of spectra per minute. Hence the might look like this.
+Say we want to define a metric for the fastest frequency for MS/MS acquisition. We look up the QC ontology and for the purpose of this document pretend, no such term exists. We coin the name `Maximal MS2 frequency` and define it as `"The fastest frequency for MS/MS collection in any minute over the complete run"`.  This will clearly be a single value type, we do not need any spectrum identifications to assign its value, but it will be for the tandem spectra of one particular run.  And it will clearly be of type frequency, in particular the sampling frequency of spectra per minute. All together, it might look like this.
 
 ```
 [Term]
@@ -130,14 +130,12 @@ relationship: has_relation QC:4000013 ! QC metric relation: one run
 ```
 where the content, context (relations), and usage are described by the is_a and relationship expressions. The unit is implied as `has_units UO:0000106 ! hertz` per the `has_relation MS:1000029 ! sampling frequency` relationship.
 
-***Issue***: do we want to allow these indirect unit associations? We can be flexible, and allow unit definition individually per object instance, or mandate. But indirect is somewhere inbetween.
-
 ## tables
 In case multiple types are involved,  the procedure changes slightly. 
 
 `Defined value type`s are no renamed unit types! In case you need multiple columns of the same type, it would not be helpful to have a `Defined value type` to be the same as a unit as this would result in column name conflicts. For this purpose, `Defined value type`s are 'named' stand-ins for unit types where same type occurrences are expected frequently. This also makes automated parsing a lot easier, as now each column has a dedicated name, and well defined/describet content, too, thanks to the defintion of the `Defined value type`.
 
-For example if we want a table to describe total ion current chromatogram forvisualisation purposes. It would be for one particular run and it would need a representation for retention time/intensity coordinates.  
+For example if we want a table to describe total ion current chromatogram for visualisation purposes. It would be for one particular run and it would need a representation for retention time/intensity coordinates.  
 We discover two types in the QC ontology which fit our needs.
 ```
 [Term]
@@ -154,7 +152,7 @@ def: "The column contains relative intensity values of one type." [PSI:QC]
 is_a: QC:4000082 ! Defined value type
 has_units: MS:1000043 ! intensity unit
 ```
-Such, we can define it as a ID free table type metric for chromatograms of a run.
+With these, we can define it as a ID free table type metric for chromatograms of a run.
 ```
 [Term]
 id: QC:4000067
--- a/cv/CV-guide.md
+++ b/cv/CV-guide.md
@@ -1,23 +1,26 @@
 # Examples of cv term use for QC
 
-## Using existing term definitions to write a 
+## Using existing term definitions to write a ...
 ### single value metric:
 If you would want to report the number of MS1 spectra for a particular run, you can use the cv Term
 ```
 [Term]
-id: QC:4000111
+id: QC:4000059
 name: Number of MS1 spectra
-def: "Number of MS1 spectra" [PSI:QC]
-comment: A lower number of MS1 spectra acquired during one sample run compared to similar runs can indicate mismatched instrument settings or issues with the instrumentation.
+def: "The number of MS1 events in the run." [PSI:QC]
 is_a: QC:4000003 ! single value
 is_a: QC:4000010 ! ID free
-is_a: QC:4000001 ! QC metric
-has_units: STATO:0000047! count
+is_a: QC:4000023 ! MS1 metric
+comment: A lower number of MS1 spectra acquired during one sample run compared to similar runs can indicate mismatched instrument settings or issues with the instrumentation or issues with sample amounts.
+relationship: has_relation MS:1000579 ! MS1 spectrum
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+property_value: has_units UO:0000189
+synonym: "MS1-Count" EXACT []
 ```
 to define your metric object which would look like 
 ```
 {
-    "accession": "QC:4000111",
+    "accession": "QC:4000059",
     "name": "Number of MS1 spectra",
     "value": 7130
 }
@@ -28,16 +31,16 @@ where the value would then be implicitly interpreted as of type `count`.
 If we choose to write a n-tuple type, for example:
 ```
 [Term]
-id: QC:4000113
+id: QC:4000138
 name: MZ acquisition range
 def: "Upper and lower limit of m/z values at which spectra are recorded." [PSI:QC]
 comment: Acquisition levels can be used as a criterion to assess the comparability of instrument settings between runs. 
 is_a: QC:4000010 ! ID free
 is_a: QC:4000001 ! QC metric
 is_a: QC:4000004 ! n-tuple
-has_units MS:1000040 ! m/z
+property_value: has_units MS:1000040
 ```
-Then we can assign the tuples as an n-array and their units implicitly assumed to be of `m/z` unit like this:
+Then we can assign the tuples as an n-array and their units implicitly assumed to be of `m/z` unit (MS:1000040) like this:
 ```
 {
     "accession": "QC:4000113",
@@ -54,41 +57,41 @@ TBD
 If we want to report multiple types of values for multiple observations, we will choose a table. A table will be defined by it's columns, which can be optional. Here is an example
 ```
 [Term]
-id: QC:4000095
+id: QC:4000120
 name: Sampling rates
-definition: "The sampling rates of identified peptides and their respective frequency (accumulated by the sampling rate, i.e. how often peptides were sampled n times and how many peptides were resampled at that rate. Optionally, which peptides were sampled at that rate, space separated)" [PSI:QC]
+def: "The sampling rates of identified peptides and their respective frequency (accumulated by the sampling rate, i.e. how often peptides were sampled n times and how many peptides were resampled at that rate. Optionally, which peptides were sampled at that rate, space separated)" [PSI:QC]
 comment: The sampling rate may give insights to peptide separation, dynamic exclusion settings and gradient efficiency.
 is_a: QC:4000001 ! QC metric
 is_a: QC:4000009 ! ID based
 is_a: QC:4000006 ! table
-has_column_type: QC:4000089 ! Sample rates
-has_column_type: QC:4000090 ! Counts
-has_optional_column_type: QC:4000091 ! Peptide sequences
+property_value: has_column QC:4000114
+property_value: has_column QC:4000115
+property_value: has_optional_column QC:4000116
 ```
 
 Any metric object of the type `Sampling rates` will be expected to contain a column of name `Sample rates`  and `Counts` and may have `Peptide sequences`. These define the respective column values as follows.
 
 ```
 [Term]
-id: QC:4000089
-name: Sample rates
-def: "The column contains sample rates as number of occurrences." [PSI:QC]
-is_a: QC:4000082 ! Defined value type
-has_units: PATO:0000161 ! rate
+id: QC:4000114
+name: Sample rate
+def: "The column contains sample rate values as number of occurrences." [PSI:QC]
+is_a: QC:4000107 ! Column type
+property_value: has_units PATO:0000161
 
 [Term]
-id: QC:4000090
-name: Counts
+id: QC:4000115
+name: Count
 def: "The column contains counts of occurrences of a common type." [PSI:QC]
-is_a: QC:4000082 ! Defined value type
-has_units: STATO:0000047 ! count
+is_a: QC:4000107 ! Column type
+property_value: has_units UO:0000189
 
 [Term]
-id: QC:4000091
-name: Peptide sequences
-def: "The column contains peptide sequences." [PSI:QC]
-is_a: QC:4000082 ! Defined value type
-has_units: NCIT:C45253 ! String
+id: QC:4000116
+name: Peptide sequence
+def: "The column contains peptide sequences; multiple peptides are separated by space." [PSI:QC]
+is_a: QC:4000107 ! Column type
+property_value: has_types NCIT:C45253
 ```
 A metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLVTHYDR") and three peptides sampled 3 times ("IQACSGNDLQQWAR","AWQQLDNGSCAQIR","MATSGMMTELGK") will look like this.
 ```
@@ -127,6 +130,8 @@ is_a: QC:4000024 ! MS2 metric
 relationship: has_relation MS:1000029 ! sampling frequency
 relationship: has_relation MS:1000580 ! MSn spectrum
 relationship: has_relation QC:4000013 ! QC metric relation: one run
+property_value: has_units UO:0000106
+synonym: "MS2-Freq-Max" EXACT []
 ```
 where the content, context (relations), and usage are described by the is_a and relationship expressions. The unit is implied as `has_units UO:0000106 ! hertz` per the `has_relation MS:1000029 ! sampling frequency` relationship.
 
@@ -139,31 +144,31 @@ For example if we want a table to describe total ion current chromatogram for vi
 We discover two types in the QC ontology which fit our needs.
 ```
 [Term]
-id: QC:4000083
-name: Retention time points
+id: QC:4000108
+name: Retention time
 def: "The column contains retention time points in seconds." [PSI:QC]
-is_a: QC:4000082 ! Defined value type
-has_units: UO:0000010 ! second
+is_a: QC:4000107 ! Column type
+property_value: has_units UO:0000010
 
 [Term]
-id: QC:4000084
-name: Relative intensities 
+id: QC:4000109
+name: Relative intensity
 def: "The column contains relative intensity values of one type." [PSI:QC]
-is_a: QC:4000082 ! Defined value type
-has_units: MS:1000043 ! intensity unit
+is_a: QC:4000107 ! Column type
+property_value: has_units MS:1000043
 ```
 With these, we can define it as a ID free table type metric for chromatograms of a run.
 ```
 [Term]
 id: QC:4000067
 name: Total ion current chromatogram
-def: "The total ion current chromatogram. Each observation is a coordinate pair of RT and intensity." [PSI:QC]
-is_a: QC:4000006 ! table
+def: "The total ion current chromatogram. The first column contains the retention time values in second, the second column the corresponding relative intensities." [PSI:QC]
+is_a: QC:4000006 ! table 
 is_a: QC:4000010 ! ID free
 is_a: QC:4000022 ! chromatogram metric
 relationship: has_relation MS:1000235 ! total ion current chromatogram
 relationship: has_relation QC:4000013 ! QC metric relation: one run
-has_column_type: QC:4000083 ! Retention time points
-has_column_type: QC:4000084 ! Relative intensities 
+property_value: has_column: QC:4000108 ! Retention time
+property_value: has_column: QC:4000109 ! Relative intensity
 ```
 
--- a/cv/CV-guide.md
+++ b/cv/CV-guide.md
@@ -1,0 +1,171 @@
+# Examples of cv term use for QC
+
+## Using existing term definitions to write a 
+### single value metric:
+If you would want to report the number of MS1 spectra for a particular run, you can use the cv Term
+```
+[Term]
+id: QC:4000111
+name: Number of MS1 spectra
+def: "Number of MS1 spectra" [PSI:QC]
+comment: A lower number of MS1 spectra acquired during one sample run compared to similar runs can indicate mismatched instrument settings or issues with the instrumentation.
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000047! count
+```
+to define your metric object which would look like 
+```
+{
+    "accession": "QC:4000111",
+    "name": "Number of MS1 spectra",
+    "value": 7130,
+}
+```
+where the value would then be implicitly interpreted as of type `count`.
+
+### n-tuple:
+If we choose to write a n-tuple type, for example:
+```
+[Term]
+id: QC:4000113
+name: MZ acquisition range
+def: "Upper and lower limit of m/z values at which spectra are recorded." [PSI:QC]
+comment: Acquisition levels can be used as a criterion to assess the comparability of instrument settings between runs. 
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+has_units MS:1000040 ! m/z
+```
+Then we can assign the tuples as an n-array and their units implicitly assumed to be of `m/z` unit like this:
+```
+{
+    "accession": "QC:4000113",
+    "name": "MZ acquisition range",
+    "value": [400,1200],
+}
+```
+
+
+### matrix:
+TBD
+
+### table:
+If we want to report multiple types of values for multiple observations, we will choose a table. A table will be defined by it's columns, which can be optional. Here is an example
+```
+[Term]
+id: QC:4000095
+name: Sampling rates
+definition: "The sampling rates of identified peptides and their respective frequency (accumulated by the sampling rate, i.e. how often peptides were sampled n times and how many peptides were resampled at that rate. Optionally, which peptides were sampled at that rate, space separated)" [PSI:QC]
+comment: The sampling rate may give insights to peptide separation, dynamic exclusion settings and gradient efficiency.
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000006 ! table
+has_column_type: QC:4000089 ! Sample rates
+has_column_type: QC:4000090 ! Counts
+has_optional_column_type: QC:4000091 ! Peptide sequences
+```
+
+Any metric object of the type `Sampling rates` will be expected to contain a column of name `Sample rates`  and `Counts` and may have `Peptide sequences`. These define the respective column values as follows.
+
+```
+[Term]
+id: QC:4000089
+name: Sample rates
+def: "The column contains sample rates as number of occurrences." [PSI:QC]
+is_a: QC:4000082 ! Defined value type
+has_units: PATO:0000161 ! rate
+
+[Term]
+id: QC:4000090
+name: Counts
+def: "The column contains counts of occurrences of a common type." [PSI:QC]
+is_a: QC:4000082 ! Defined value type
+has_units: STATO:0000047 ! count
+
+[Term]
+id: QC:4000091
+name: Peptide sequences
+def: "The column contains peptide sequences." [PSI:QC]
+is_a: QC:4000082 ! Defined value type
+has_units: NCIT:C45253 ! String
+```
+Such, a metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLVTHYDR") and three peptides sampled 3 times ("IQACSGNDLQQWAR","AWQQLDNGSCAQIR","MATSGMMTELGK") will look like this.
+```
+{
+    "accession": "QC:4000113",
+    "name": "MZ acquisition range",
+    "value": {
+	    "Sample rates": [5,3],
+	    "Counts": [2,3],
+	    "Peptide sequences": ["DYHTVLGAR AGLVTHYDR","IQACSGNDLQQWAR AWQQLDNGSCAQIR MATSGMMTELGK"],
+    },
+}
+```
+
+
+# Define a new metric term for a
+If a desired metric is not yet defined, you will need a cv term definition before you can write a metric object. 
+
+## single value, n-tuple, matrix:
+This is straight forward for these types. 
+1. Make sure, the term is not defined yet in the QC ontology
+2. Define a reasonable descriptive name while keeping concise
+3. Describe the term content, context (relations), and usage 
+4. Define its value type 
+
+Say we want to define a metric for the fastest frequency for MS/MS acquisition. We look up the QC ontology and for the purpose of this document pretend, no such term exists. We coin the name `Maximal MS2 frequency` and define it as `"The fastest frequency for MS/MS collection in any minute over the complete run"`.  This will clearly be a single value type, we do not need any spectrum identifications to assign its value, but it will be for the MS2 tandem spectra of one particular run.  And it will clearly be of type frequency, in particular the sampling frequency of spectra per minute. Hence the might look like this.
+
+```
+[Term]
+id: QC:4000061
+name: Maximal MS2 frequency
+def: "The fastest frequency for MS/MS collection in any minute over the complete run" [PSI:QC]
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000024 ! MS2 metric
+relationship: has_relation MS:1000029 ! sampling frequency
+relationship: has_relation MS:1000580 ! MSn spectrum
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+```
+where the content, context (relations), and usage are described by the is_a and relationship expressions. The unit is implied as `has_units UO:0000106 ! hertz` per the `has_relation MS:1000029 ! sampling frequency` relationship.
+
+***Issue***: do we want to allow these indirect unit associations? We can be flexible, and allow unit definition individually per object instance, or mandate. But indirect is somewhere inbetween.
+
+## tables
+In case multiple types are involved,  the procedure changes slightly. 
+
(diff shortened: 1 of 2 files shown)


cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+id: QC:4000076
+name: Length values
+def: "The column contains length values of sequences, i.e. the count of items within the sequences, e.g. total length of proteins." [PSI:QC]
+is_a: QC:4000071 ! Column type
+has_units: STATO:0000047 ! count
+
+[Term]
+id: QC:4000077
+name: Target or decoy designation
+def: "The column contains designations of the type 'target' or 'decoy'." [PSI:QC]
```
cbielow, line 465: what about "target+decoy"?

cbielow: if it matches both

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+id: QC:4000079
+name: Counts
+def: "The column contains counts of occurrences of a common type." [PSI:QC]
+is_a: QC:4000071 ! Column type
+has_units: STATO:0000047 ! count
+
+[Term]
+id: QC:4000080
+name: Peptide sequences
+def: "The column contains peptide sequences." [PSI:QC]
```
cbielow, line 486: can sequences carry modifications?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000006 ! table
+has_column: QC:4000078 ! Sample rates
+has_column: QC:4000079 ! Counts
+has_optional_column: QC:4000080 ! Peptide sequences
+
+[Term]
+id: QC:40000
+name: Explained precursor intensity first quarter
```
cbielow, line 531: this description is very confusing....
what exactly is computed here?
Of all MS2 scans whose precursor intensity is within the first quarter of all precursor intensities, how many were identified?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+comment: Can be used to approximate the dynamic range of signal
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000300 ! dimensionless ratio
+synonym: "MS1-3A" NARROW []
+
+[Term]
+id: QC:40000
+name: Precursor intensity range of identified MS2
```
cbielow, line 582: there is `Precursor intensity of identified MS2 Q1, Q2, Q3` and `Precursor intensity of unidentified MS2 Q1, Q2, Q3`, but there is no `Precursor intensity range of unidentified MS2`

cv/qc-cv.obo
mwalzer: good catch

cv/qc-cv.obo
mwalzer: I wouldn't want to restrict the contents too much. Would this likely break 'metric consumer code' if we'd say sequences in any common representation format?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of precursor intensity of unidentified MS2, the quartiles Q1, Q2, Q3" [PSI:QC] 
+comment: The (un)identified precursor intensity distribution can aid the interpretation of overall identification success.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+has_units: STATO:0000167 ! first quartile, STATO:0000574 ! center value, STATO:0000170 ! third quartile
+
+[Term]
+id: QC:40000
+name: Median S/N for MS1 spectra in RT range in which half of the peptides are identified
```
cbielow, line 612: how is this 'half' computed? There are a lot of RT ranges which fulfill this property....

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:40000
+name: Charged spectra ratio +1 over +2
+def: Ratio of 1+ spectra count over 2+ spectra count in all MS2
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+synonym: "IS-3A"  RELATED []
```
cbielow, line 1057: add this to a 
```remark:
    synonyms can be 'RELATED` (the metric is similar, but not the same thing) or 'NARROW' (identical interpretation of numbers)
```
section on top of this .obo... so people know how to read this

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+name: Charged spectra ratio +4 over +2
+def: Ratio of 4+ peptide count over 2+ peptide count in all MS2
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+synonym: "IS-3C"  RELATED []
+
+[Term]
+id: QC:40000
+name: Mean charge in all MS2
```
cbielow, line 1079: mean of all precursor (!) charge states across all MS2? (including unidentified?)

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: Mean charge in all MS2
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:40000
+name: Median charge in all MS2
```
cbielow, line 1087: same issue as above

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: Median charge in all MS2
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:40000
+name: Number of missed cleavages
```
cbielow, line 1095: sum of all missed cleavages across all identifications? Does this make any sense?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: Number of missed cleavages
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000047! count
+
+[Term]
+id: QC:40000
+name: Total number of protein evidences
```
cbielow, line 1104: what is a protein evidence? A PSM, a peptide?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: Total number of protein evidences
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000047! count
+
+[Term]
+id: QC:40000
+name: Number of identified proteins
```
cbielow, line 1113: at what FDR?

cbielow: using which algorithm?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+name: ID ratio
+def: The ratio of identified and recorded spectra.
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000300 ! dimensionless ratio
+
+[Term]
+id: QC:40000
+name: Precursor errors (abs) Q1, Q2, Q3
```
cbielow, line 1158: what is the unit of error?
Da or PPM?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of Precursor errors (abs), the quartiles Q1, Q2, Q3 value" [PSI:QC]
+comment: The absolute precursor error distribution can aid the interpretation of overall identification success.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+has_units: STATO:0000167 ! first quartile, STATO:0000574 ! center value, STATO:0000170 ! third quartile
+
+[Term]
+id: QC:40000
+name: Precursor errors (abs) mean
```
cbielow, line 1168: same as above...

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000401 ! sample mean
+synonym: "MS1-5B" NARROW []
+
+[Term]
+id: QC:40000
+name: Precursor errors (abs) sigma
+def: "From the distribution of Precursor errors (abs), the sigma value" [PSI:QC]
```
cbielow, line 1180: assuming a Gaussian? or what is Sigma?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+comment: The absolute precursor error distribution can aid the interpretation of overall identification success.
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000237 ! standard deviation
+
+[Term]
+id: QC:40000
+name: Precursor errors (abs) outliers (<Q1-1.5*IQR, >Q3+1.5*IQR)
```
cbielow, line 1189: what is the unit?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+comment: The absolute precursor error distribution can aid the interpretation of overall identification success.
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000006 ! table
+has_column: QC:4000081 ! Outliers below
+has_column: QC:4000082 ! Outliers above
+
+[Term]
+id: QC:40000
+name: Precursor errors (ppm) Q1, Q2, Q3
```
cbielow, line 1200: aha... ppm...
so the others above are Da?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of Precursor errors (ppm), the list of outliers below of Q1-1.5IQR and above Q3+1.5IQR" [PSI:QC]
+comment: The precursor error in ppm distribution can aid the interpretation of overall identification success.
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000006 ! table
+has_column: QC:4000081 ! Outliers below
+has_column: QC:4000082 ! Outliers above
+
+[Term]
+id: QC:40000
```
cbielow, line 1240: can you move that metric up, so its in order?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000401 ! sample mean
+synonym: "MS1-5D" NARROW []
+
+[Term]
+id: QC:40000
+name: Identification score Q1, Q2, Q3
+def: "From the distribution of Identification score, the Q1, Q2, Q3 value" [PSI:QC]
+comment: The precursor error in ppm distribution can aid the interpretation of overall identification success.
```
cbielow, line 1276: comment does not fit

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of Identification score, the mean value" [PSI:QC]
+comment: The identification score distribution can help improve scoring function tuning, aid the interpretation of overall identification success, and integration with downstream processing such as percolator.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000003 ! single value
+has_units: STATO:0000401 ! sample mean
+
+[Term]
+id: QC:40000
+name: Identification score sigma
```
cbielow, line 1294: does sigma make any sense for many of the scores out there?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of identified peptide lengths the quartiles Q1, Q2, Q3 value" [PSI:QC]
+comment: The identified peptide lengths distribution can aid the interpretation of digestion efficacy.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+has_units: STATO:0000167 ! first quartile, STATO:0000574 ! center value, STATO:0000170 ! third quartile
+
+[Term]
+id: QC:40000
+name: Missed cleavage mean
```
cbielow, line 1367: duplicate (see above)

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+comment: The identified peptide lengths distribution can aid the interpretation of digestion efficacy.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000003 ! single value
+has_units: STATO:0000401 ! sample mean
+synonym: "MS1-5B" NARROW []
+
+[Term]
+id: QC:40000
+name: Missed cleavage sigma
```
cbielow, line 1378: duplicate (see above)

.... maybe the OBO should be run through a script to detect plain duplicates?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of identified peptide lengths the sigma value" [PSI:QC]
+comment: The identified peptide lengths distribution can aid the interpretation of digestion efficacy.
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000003 ! single value
+has_units: STATO:0000237 ! standard deviation
+
+[Term]
+id: QC:40000
+name: Missed cleavage outliers (<Q1-1.5*IQR, >Q3+1.5*IQR)
```
cbielow, line 1388: duplicate

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+comment: The identified peptide lengths distribution can aid the interpretation of digestion efficacy.
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000006 ! table
+has_column: QC:4000081 ! Outliers below
+has_column: QC:4000082 ! Outliers above
+
+[Term]
+id: QC:40000
+name: Signal-to-noise ratio in MS1 Q1, Q2, Q3
```
cbielow, line 1399: is this computed on raw or profile data? And then one S/N value per data point?
And how is S/N determined?

cbielow: if the algorithm for S/N is deliberately not mentioned, say that the user can select whatever he likes, but the results are not comparable.

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of signal-to-noise ratio in MS1, the sigma value" [PSI:QC]
+comment: The Signal-to-noise ratio in MS1 distribution can provide insight into the presence of nuisance factors.
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000237 ! standard deviation
+
+[Term]
+id: QC:40000
+name: Signal-to-noise ratio in MS1 outliers (<Q1-1.5*IQR, >Q3+1.5*IQR)
```
cbielow, line 1429: what are MS1 outliers?

cv/qc-cv.obo
```diff
@@ -417,6 +417,1146 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "From the distribution of signal-to-noise ratio in MS2, the mean" [PSI:QC]
+comment: A high signal-to-noise ratio in MS2 distribution can explain a high rate of unidentified spectra.
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000401 ! sample mean
+
+[Term]
+id: QC:40000
+name: Signal-to-noise ratio in MS1 sigma
```
cbielow, line 1460: duplicate
Did you mean MS2 sigma?

cv/qc-cv.obo
mwalzer: was copy-and-paste oversight (applies to comments above, too)

cv/qc-cv.obo
mwalzer: Same as with FDR for the identification numbers. there is a place for the software in mzQC, which can also detail algorithms used I presume. I'll add a comment for interpretation aid.

cv/qc-cv.obo
mwalzer: 'Signal-to-noise ratio in MS1' outliers

cv/qc-cv.obo
mwalzer: yep, it is correct in the definition.

cv/qc-cv.obo
mwalzer: I didn't add that as it will probably not be that informative; there is going to be unidentified all over the int range.

cv/qc-cv.obo
mwalzer: True, there is a metric later that uses the shortest RT period with 50%. This here is Q1 to Q3, I'll add more descr.

cv/qc-cv.obo
mwalzer: :+1:

cv/qc-cv.obo
mwalzer: yep, all as in all (is_a: QC:4000010 ! ID free)

cv/qc-cv.obo
mwalzer: indeed, we have the mc distribution metrics later which should be informative, probably a template remnant.

cv/qc-cv.obo
mwalzer: :)

cv/qc-cv.obo
mwalzer: Valid point, I'll add a comment. FDR threshold might be interesting though. Maybe a tuple instead of a single value

cv/qc-cv.obo
mwalzer: see definition

cv/qc-cv.obo
```diff
@@ -6,12 +6,13 @@ default-namespace: QC
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 namespace-id-rule: * QC:$sequence(7,0,9999999)$
 remark: coverage: Mass spectrometer quality control metrics
-remark: creator: Chris Bielow <chris.bielow <-at-> fu-berlin.de>
 remark: creator: Julian Uszkoreit <julian.uszkoreit <-at-> ruhr-uni-bochum.de>
-remark: creator: Martin Eisenacher <martin.eisenacher <-at-> ruhr-uni-bochum.de>
+remark: synonyms can be 'RELATED' (the metric is similar, but not the same thing), 'NARROW' (identical interpretation of numbers), 'EXACT' (result of renaming)
```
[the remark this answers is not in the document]
mwalzer: very good point. I'll add explanations in the remarks and the spec doc

cv/qc-cv.obo
```diff
@@ -425,6 +1559,14 @@ name: has_relation
 id: has_type
 name: has_type
 
+[Typedef]
+id: has_column
+name: has_column
+
```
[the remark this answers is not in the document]
mwalzer: for matrix I don't think so. I added the other value types and some comments. Are comments in that place of the obo legal?

cv/qc-cv.obo
```diff
@@ -417,6 +418,1139 @@ is_a: QC:4000024 ! MS2 metric
...
+def: "The parent to the group of column definitions for name, content, and units for table type." [PSI:QC]
+relationship: has_relation id: QC:4000006 ! table
+
+[Term]
+id: QC:4000083
+name: Retention times
+def: "The column contains retention times in seconds." [PSI:QC]
+is_a: QC:4000082 ! Column type
+has_units: UO:0000010 ! second
+
```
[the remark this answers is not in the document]
mwalzer: I used time as in time points, but just lazily omitting the point. Could be more concise, I conceed.

cv/qc-cv.obo
```diff
@@ -417,6 +418,1139 @@ is_a: QC:4000024 ! MS2 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 relationship: has_relation QC:4000067 ! Total ion current chromatogram
 
+[Term]
+id: QC:4000082
+name: Column type
```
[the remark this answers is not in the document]
mwalzer: k, changed it.

cv/qc-cv.obo
```diff
@@ -417,6 +418,1139 @@ is_a: QC:4000024 ! MS2 metric
 relationship: has_relation QC:4000013 ! QC metric relation: one run
 relationship: has_relation QC:4000067 ! Total ion current chromatogram
```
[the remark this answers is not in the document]
mwalzer: Mmmh, should we get rid of corresponding lists? They are in essence tables...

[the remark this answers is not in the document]
cbielow: > 
> 
> Mmmh, should we get rid of corresponding lists? They are in essence tables...

+1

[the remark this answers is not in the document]
julianu: Right, only without "column names". But I guess we can live without them.
So yes, remove them.

cv/qc-cv.obo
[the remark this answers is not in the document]
julianu: It's not only the retention times, but also all other "QC:4000082 ! Column type"s.

cv/qc-cv.obo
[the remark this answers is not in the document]
julianu: Where did you add the comments?
I guess thereis a way though how to add comments, maybe the "remark" tag?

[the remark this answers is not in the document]
mwalzer: So 'comments' for now just in the [term] stanzas. Examples:
https://github.com/HUPO-PSI/mzQC/pull/102/files#diff-a8418b759125587711768b57da054b1eR65
https://github.com/HUPO-PSI/mzQC/pull/102/files#diff-a8418b759125587711768b57da054b1eR389
So should I try 'remark' for [typedef]?

cv/qc-cv.obo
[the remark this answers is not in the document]
mwalzer: corresponding_list is removed, ... about the types I refer to the discussion about n-tuples we'll have on friday.

cv/qc-cv.obo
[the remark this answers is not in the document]
mwalzer: About typedef consistency: do you prefer `has_value_type`, `has_column_type`, `has_tuple_type` or rather `has_type`, `has_column`, `has_tuple`?

mwalzer:
@julianu Can you have a final look over the added terms? We wanted to do some changes in other PRs based on (this) latest version of the .obo if I remember correctly.

cv/qc-cv.obo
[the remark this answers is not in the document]
mwalzer: K, changed all to singular

cv/qc-cv.obo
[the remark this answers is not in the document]
mwalzer: I guess the latter is alright now?

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Using existing term definitions to write a 
+### single value metric:
+If you would want to report the number of MS1 spectra for a particular run, you can use the cv Term
+```
+[Term]
+id: QC:4000111
+name: Number of MS1 spectra
+def: "Number of MS1 spectra" [PSI:QC]
+comment: A lower number of MS1 spectra acquired during one sample run compared to similar runs can indicate mismatched instrument settings or issues with the instrumentation.
```
cbielow, line 11: or issues with sample amount ...

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+has_units: STATO:0000047! count
+```
+to define your metric object which would look like 
+```
+{
+    "accession": "QC:4000111",
+    "name": "Number of MS1 spectra",
+    "value": 7130,
```
cbielow, line 22: the trailing `,` is technically not valid in JSON

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+### matrix:
+TBD
+
+### table:
+If we want to report multiple types of values for multiple observations, we will choose a table. A table will be defined by it's columns, which can be optional. Here is an example
+```
+[Term]
+id: QC:4000095
+name: Sampling rates
+definition: "The sampling rates of identified peptides and their respective frequency (accumulated by the sampling rate, i.e. how often peptides were sampled n times and how many peptides were resampled at that rate. Optionally, which peptides were sampled at that rate, space separated)" [PSI:QC]
```
cbielow, line 59: I dont understand this definition...
If its a `rate` it should be a rational number, but below, its simply counts.

And how can a peptide be sampled, e.g., 5 times, and that occurs 3 times? Is this metric accumulating data over many  runs?

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+id: QC:4000090
+name: Counts
+def: "The column contains counts of occurrences of a common type." [PSI:QC]
+is_a: QC:4000082 ! Defined value type
+has_units: STATO:0000047 ! count
+
+[Term]
+id: QC:4000091
+name: Peptide sequences
+def: "The column contains peptide sequences." [PSI:QC]
```
cbielow, line 89: `. Multiple peptides can be separated by space`

cv/CV-guide.md
cbielow: ok, now I get it.... sorry.
is this more clear: `In this table, each row contains three values: Sample rate 's', i.e. how often a peptide was sampled in a run; a count 'c', for how many peptides were sampled this often ('s' times); peptide sequences (a single string of exactly 'c' peptides). The sampling rate must be unique across all rows.`??
This is mainly about the CV term, not the PR per se though...

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+is_a: QC:4000082 ! Defined value type
+has_units: NCIT:C45253 ! String
+```
+Such, a metric object for two peptides being samples 5 times ("DYHTVLGAR", "AGLVTHYDR") and three peptides sampled 3 times ("IQACSGNDLQQWAR","AWQQLDNGSCAQIR","MATSGMMTELGK") will look like this.
```
cbielow, line 93: Remove `Such, `?
Maybe `Thus, `?

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+    "value": {
+	    "Sample rates": [5,3],
+	    "Counts": [2,3],
+	    "Peptide sequences": ["DYHTVLGAR AGLVTHYDR","IQACSGNDLQQWAR AWQQLDNGSCAQIR MATSGMMTELGK"],
+    },
+}
+```
+
+
+# Define a new metric term for a
```
cbielow, line 107: ` # Define a new metric term for a ...` 
?

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+If a desired metric is not yet defined, you will need a cv term definition before you can write a metric object. 
+
+## single value, n-tuple, matrix:
+This is straight forward for these types. 
+1. Make sure, the term is not defined yet in the QC ontology
+2. Define a reasonable descriptive name while keeping concise
+3. Describe the term content, context (relations), and usage 
+4. Define its value type
```
cbielow, line 115: (and unit)

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+If a desired metric is not yet defined, you will need a cv term definition before you can write a metric object. 
+
+## single value, n-tuple, matrix:
+This is straight forward for these types. 
+1. Make sure, the term is not defined yet in the QC ontology
+2. Define a reasonable descriptive name while keeping concise
+3. Describe the term content, context (relations), and usage 
+4. Define its value type 
+
+Say we want to define a metric for the fastest frequency for MS/MS acquisition. We look up the QC ontology and for the purpose of this document pretend, no such term exists. We coin the name `Maximal MS2 frequency` and define it as `"The fastest frequency for MS/MS collection in any minute over the complete run"`.  This will clearly be a single value type, we do not need any spectrum identifications to assign its value, but it will be for the MS2 tandem spectra of one particular run.  And it will clearly be of type frequency, in particular the sampling frequency of spectra per minute. Hence the might look like this.
```
cbielow, line 117: `MS2 tandem spectra ` seems a bit duplicated

`Hence the might look like this.` .. --> `it`?

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+```
+where the content, context (relations), and usage are described by the is_a and relationship expressions. The unit is implied as `has_units UO:0000106 ! hertz` per the `has_relation MS:1000029 ! sampling frequency` relationship.
+
+***Issue***: do we want to allow these indirect unit associations? We can be flexible, and allow unit definition individually per object instance, or mandate. But indirect is somewhere inbetween.
+
+## tables
+In case multiple types are involved,  the procedure changes slightly. 
+
+`Defined value type`s are no renamed unit types! In case you need multiple columns of the same type, it would not be helpful to have a `Defined value type` to be the same as a unit as this would result in column name conflicts. For this purpose, `Defined value type`s are 'named' stand-ins for unit types where same type occurrences are expected frequently. This also makes automated parsing a lot easier, as now each column has a dedicated name, and well defined/describet content, too, thanks to the defintion of the `Defined value type`.
```
cbielow, line 138: `Defined value type` comes a bit out of nowhere.
Mention that its a CV term: `QC:4000082`.
But otherwise I'm not sure what this paragraph wants to say.

How about: `Each column in your table requires a separate CV term, which is derived from 'Defined value type' (QC:4000082), e.g. one column could be 'Sampling rate' (QC:4000089), another one a 'Count' (...).
This ensures that every column has a unique name (thus order is not important) and well defined content.`

cv/CV-guide.md
```diff
@@ -0,0 +1,171 @@
...
+
+For example if we want a table to describe total ion current chromatogram forvisualisation purposes. It would be for one particular run and it would need a representation for retention time/intensity coordinates.
```
cbielow, line 140: `for visualization`

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+id: QC:4000109
+name: Relative intensity
+def: "The column contains relative intensity values of one type." [PSI:QC]
+is_a: QC:4000107 ! Column type
+property_value: has_units MS:1000043
+
+[Term]
+id: QC:4000110
+name: Sequence coverage
+def: "The column contains sequence coverage values in percent, e.g. from proteins." [PSI:QC]
```
cbielow, line 590: make it explicit that allowed values are in range [0,100]?
Otherwise people might use [0,1]

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+id: QC:4000108
+name: Retention time
+def: "The column contains retention time points in seconds." [PSI:QC]
+is_a: QC:4000107 ! Column type
+property_value: has_units UO:0000010
+
+[Term]
+id: QC:4000109
+name: Relative intensity
+def: "The column contains relative intensity values of one type." [PSI:QC]
```
cbielow, line 583: what is meant by `one type`?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+property_value: has_units STATO:0000167
+property_value: has_units STATO:0000574
+property_value: has_units STATO:0000170
+
+[Term]
+id: QC:4000129
+name: Median S/N for MS1 spectra in RT range in which the middle half of the peptides are identified
```
cbielow, line 763: puhh.. thats a mouthful.

and is S/N based on intensity?
`middle half` is also weird...

`Median S/N of MS1 spectra whose RT is between Q1-Q3 of identified MS2`?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+comment: The intensity range of the precursors informs about the dynamic range of the acquisition. 
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+property_value: has_units MS:1003085
+
+[Term]
+id: QC:4000145
+name: MS1 frequency in RT quarters
+def: "MS1 frequency in RT 1st quarter, 2nd quarter, 3rd quarter, 4th quarter" [PSI:QC]
```
cbielow, line 918: so quarter != quantile, right?
This is a bit confusing why quarter is used here.
Is this implemented somewhere?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+def: "MS2 frequency in RT 1st quarter, 2nd quarter, 3rd quarter, 4th quarter" [PSI:QC]
+comment: Spectrum acquisition frequency can be used to gauge the suitability of used instrument settings for the sample content used.
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000004 ! n-tuple
+property_value: has_units UO:0000106
+
+[Term]
+id: QC:4000147
+name: MS1 trap time Q1, Q2, Q3
```
cbielow, line 937: can this be made a bit more 'wide' in scope?
Do TOF's have something similar?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+comment: The intensity distribution of the precursors informs about the dynamic range of the acquisition.
+is_a: QC:4000001 ! QC metric
+is_a: QC:4000009 ! ID based
+is_a: QC:4000004 ! n-tuple
+property_value: has_column QC:4000118
+
+[Term]
+id: QC:4000172
+name: MS1 signal jump (10x) count
+def: "The count of MS1 signal jump by a factor of ten or more (10x) between two subsequent scans" [PSI:QC]
```
cbielow, line 1196: the jump is quantified how? Maximum intensity or mean or ...

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+name: Median precursor charge in all MS2
+def: "Median precursor charge in all MS2" [PSI:QC]
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+
+[Term]
+id: QC:4000184
+name: Number of different proteins from all PSM
+def: "Number of different protein from all PSM" [PSI:QC]
```
cbielow, line 1302: mhh.... does this include indistinguishable proteins, e.g. from a group where you don't know which one is present?
or only 'majority proteins' (as MQ would call them)?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+def: "Number of identified proteins at given FDR threshold, first number is the number of proteins (considering sequence only), second number is the FDR threshold applied (negative if no threshold applied)" [PSI:QC]
+comment: The number of identifications are likely to only make local sense within a confined set of identification parameters (identification algorithm, searchspace, FDR method).
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+property_value: has_units UO:0000189
+
+[Term]
+id: QC:4000186
+name: Total number of PSM
```
cbielow, line 1320: of PSMs

cbielow: before or after FDR?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+def: "Number of identified spectra at given FDR threshold, first number is the number of spectra, second number is the FDR threshold applied (negative if no threshold applied)" [PSI:QC]
+comment: The number of identifications are likely to only make local sense within a confined set of identification parameters (identification algorithm, searchspace, FDR method).
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+property_value: has_units UO:0000189
+
+[Term]
+id: QC:4000189
+name: ID ratio
```
cbielow, line 1349: after FDR?

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+def: "The ratio of identified and recorded spectra." [PSI:QC]
+is_a: QC:4000003 ! single value
+is_a: QC:4000009 ! ID based
+is_a: QC:4000001 ! QC metric
+property_value: has_units STATO:0000300
+
+[Term]
+id: QC:4000190
+name: Precursor errors (Da) Q1, Q2, Q3
```
cbielow, line 1358: there are at least two ways how this could be interpreted:
1) mass deviation of PC to identified peptide
2) mass deviation of PC when just correcting the isotope peak its pointing to (without doing any ID).

there is probably more ways...

cv/qc-cv.obo
```diff
@@ -539,14 +564,1316 @@ is_a: QC:4000009 ! ID based
...
+comment: The signal-to-noise ratio in MS1 distribution can provide insight into the presence of nuisance factors. Comparisons will likely only make local sense where the same method of S/N calculation is applied (or overall the same software to calculate this metric, see mzQC 'analysisSoftware').
+def: "From the distribution of signal-to-noise ratio in MS1, the mean" [PSI:QC]
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000001 ! QC metric
+property_value: has_units STATO:0000401
+
+[Term]
+id: QC:4000220
+name: Signal-to-noise ratio in MS1 - sigma
```
cbielow, line 1676: I like this `-`. Can this be used everywhere?

cbielow (changes requested):
added my 50ct :)

cv/qc-cv.obo
mwalzer: I changed it (to distinct) and added one term for indistinguishable proteins

cv/qc-cv.obo
mwalzer: this is before, there is a term for 'after given threshold' filtering to each spectra, peptides, proteins

cv/qc-cv.obo
mwalzer: so there is a set of terms for ppm mass difference (identified and precursor), so I think this should be mass difference in Da just to balance.

mwalzer: I made changes there

cv/qc-cv.obo
mwalzer: Made changes there.

cv/qc-cv.obo
mwalzer: Made changes there.

cv/qc-cv.obo
mwalzer: the unit is a family term with [multiple](https://www.ebi.ac.uk/ols/ontologies/ms/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMS_1000043&viewMode=All&siblings=false), but the column is supposed to only contain  values of one type.

cv/CV-guide.md
mwalzer: As soon as the term is fixed in the obo, this'll get an update

cv/CV-guide.md
mwalzer: As soon as the term is fixed in the obo, this'll get an update

cv/qc-cv.obo
cbielow: `Minimal set of proteins, which explains all available evidence (PSMs)`

cv/qc-cv.obo
```diff
@@ -9,12 +9,38 @@ remark: coverage: Mass spectrometer quality control metrics
 remark: creator: Chris Bielow <chris.bielow <-at-> fu-berlin.de>
 remark: creator: Julian Uszkoreit <julian.uszkoreit <-at-> ruhr-uni-bochum.de>
 remark: creator: Martin Eisenacher <martin.eisenacher <-at-> ruhr-uni-bochum.de>
+remark: creator: Mathias Walzer <walzer <-at-> ebi.ac.uk>
 remark: namespace: MS
 remark: namespace: QC
 import: http://ontologies.berkeleybop.org/uo.obo
 import: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
```
cbielow, line 16: STATO would need an import here, right?

cbielow: never mind. Done in #113

Issue #11, opened 2015-03-17
extend sample files and datasets


GoogleCodeExporter:
```
A extended sample file(s) comprising all features of qcML
is needed for implementation aid and testing.
```

Original issue reported on code.google.com by `mathias....@gmail.com` on 6 Sep 2013 at 8:08

julianu:
#4

bittremieux:
Example files to produce:

- [ ] single input file (QC-Calculator @mwalzer)
- [ ] multiple input files (PTXQC @cbielow: derived from MaxQuant's proteins.txt file).
- [ ] metabolomics (@rsalek)

Possible extra files?

- [ ] QuaMeter (@dtabb73)
- [ ] Yamato / SwaMe (@PaulBrack)

dtabb73:
I have requested that Marina Kriek produce an mzQC from her updated build of QuaMeter for the 120-experiment Heunis outlier set.

bittremieux:
mzQC version of the 120-experiment outlier set [here](https://gist.github.com/bittremieux/a3a43ce929e38fd0230005782f3c9725).

Still missing: units of the metrics.

bittremieux:
We have the essential example files now, so closing. Of course, additional relevant examples can always be added.

PR #110, opened 2020-12-03
adding optional fields in schema

mwalzer:
* on top-level: readMe, contanctName, contactAddress
* metadata: labelName


schema/mzqc_schema.json
```diff
@@ -23,6 +23,14 @@
                     "description": "Description and comments about the mzQC file contents.",
                     "type": "string"
                 },
+                "contactName": {
+                    "description": "Name of contact a particular mzQC file.",
+                    "type": "string"
+                },
+                "contactAddress": {
+                    "description": "Contact Address (mail/tel.) for getting in touch with given contact for a particular file",
```
cbielow, line 31: `particular mzQC file`

schema/mzqc_schema.json
```diff
@@ -23,6 +23,14 @@
                     "description": "Description and comments about the mzQC file contents.",
                     "type": "string"
                 },
+                "contactName": {
+                    "description": "Name of contact a particular mzQC file.",
```
cbielow, line 27: `Creator of this mzQC file`

schema/mzqc_schema.json
```diff
@@ -147,6 +159,10 @@
                         ]
                     }
                 },
+                "labelName": {
+                    "description": "OPTIONAL label name. For setQuality, this a group name, lending itself for example as a axis labels for a plot. OPTIONAL.",
```
cbielow, line 163: we need to discuss how this relates to inputFile:name

schema/mzqc_schema.json
mwalzer: Not necessarily creator?

schema/mzqc_schema.json
mwalzer: :+1:

schema/mzqc_schema.json
mwalzer: I'd suggest to describe it in the spec doc with something like "We suggest to use labelName only if either a label different from the inputFile.name needs to be specified or grouping labels are wanted. In case references for a setQuality are needed, labelName needs to be used."

schema/mzqc_schema.json
bittremieux: I agree, it can be the PI as well (and probably should for long-term availability).

schema/mzqc_schema.json
```diff
@@ -19,6 +19,18 @@
                     "type": "string",
                     "format": "date-time"
                 },
+                "readMe": {
```
bittremieux, line 22: I'd maybe prefer "description" over "readMe" because that feels a bit less like an obligation imo.

schema/mzqc_schema.json
bittremieux: Maybe "label" rather than "labelName". The latter is kind of two times the same thing?

bittremieux (commented):
I just have some minor naming comments.

I remember that we had contact information in a qcML version way back. Apparently, just like comments, those didn't survive the transition. I don't remember why not though.

schema/mzqc_schema.json
mwalzer: true :smile: dubbelop  
`label` then?

schema/mzqc_schema.json
mwalzer: Ok

bittremieux:
TODO:

[ ] Remove `fileProvenance`.

mwalzer:
takes care of #11 with up-to-date examples in v1.0

Issue #111, opened 2020-12-22
example renaming


cbielow:
We should discuss the current naming scheme for the examples in `./doc/`:
`single-run.mzQC` describes a single run, but essentially its about 'runQualities'. So its easy to store multiple, but independent runs.

However, the name of `multi-run.mzQC` suggests that it explains how to store multiple runs, but in fact it explains `setQualities`.

I'd suggest to use different names:
`individual-runs.mzQC` and `set-of-runs.mzQC`
Other suggestions welcome :)
I know this somewhat breaks the links in the mzQC spec doc, but it should not be hard to adapt those before release 1.0.0, right?

mwalzer:
Do you want to add this to #112 ?

cbielow:
jep. can do

PR #112, opened 2020-12-22
mzQC example files for setQuality (and runQuality)

cbielow:
- augmented the runQuality example
- added example for setQuality

I tried to make it as easy as possible.
We should discuss, if this `superset` solution is a good way to go, or if we can think of something better...
E.g. this might not work if you want to combine multiple mzQC files into one document and the names for the supersets are not unique across files.


bittremieux:
What is actually the "superset" in this example? Is it not rather the "set label"?

cbielow:
the superset is the matrix of PCA components (PC dimensions per each set).

cbielow:
image

This image explains it best, maybe: each dot in represents a group. Thus, one setQuality contributes one dot (represented a using an array).

cbielow:
can we go ahead here?

mwalzer:
Could the PCA plot be integrated into the .md to explain more visually?

mwalzer:
The way PCA is represented is a bit confusing to me. So usually, when you do a PCA analysis, you'll put in some selected features for each of your data points, say a run. But you'll enter them at once. And you'll get all the PCs for each data point in return, basically one set. From the image, I assume that the example should have 5 inputs, each sets? We have a different example that also deals with PCA in the metabo-batches.mzQC
Do you need to report each set PCA individually and then plot the PCA combined?

doc/multi-run.mzQC.md
```diff
@@ -0,0 +1,310 @@
...
+    "cvParameters": [
+      {
+        "accession": "QC:0000000",
+        "name": "superset-label",
+        "value": "MaxQuant experiment groups"
+      }
+    ]
+  }
+```
+In other words: if you need to represent some kind of hierarchy of QC information, maybe due to your experimental design, `setQuality`'s are always the leaf nodes in that tree. To describe data for inner nodes, annotate the sets in the leaf nodes with a CV term. This allows to reconstruct this hierarchy.
```
mwalzer, line 121: Maybe we could build up one-by-one, first explaining set and label, then super-sets and other metadata cv parameters?

mwalzer (changes requested):

cbielow:
> 
> 
> Could the PCA plot be integrated into the .md to explain more visually?

done

cbielow:
> 
> 
> The way PCA is represented is a bit confusing to me. So usually, when you do a PCA analysis, you'll put in some selected features for each of your data points, say a run. But you'll enter them at once. And you'll get all the PCs for each data point in return, basically one set. From the image, I assume that the example should have 5 inputs, each sets? We have a different example that also deals with PCA in the metabo-batches.mzQC
> Do you need to report each set PCA individually and then plot the PCA combined?

Yes. I see no other way.

cbielow:
Ok, the example is brushed up, with an image and the suggestions by Mathias.
The files are also renamed, as suggested in #111.

mwalzer:
@bittremieux merge okay from your end?

mwalzer:
We need to update the CV terms in a second pass! GitHub doesn't have a reminder function built-in yet, eh?

doc/individual-runs.mzQC.md
```diff
@@ -30,34 +40,36 @@ cv term reference information at the bottom,
...
+            ...
         },
         "qualityMetrics": [
-...
+            ...
         ]
       }
 ```
-The inputFiles consist of a single object, describing the source file with structured information about the file's name, format, location and other properties, defined via cv terms. 
+The `inputFiles` consist of a single object, describing the source file with structured information about the file's name, format, location and other properties, defined via cv terms.
```
bittremieux, line 67: Difference between the list of input files (`inputFiles`) and the individual elements in there?

doc/set-of-runs.mzQC.md
```diff
@@ -0,0 +1,312 @@
...
+In  `qualityMetrics`, we will store the actual QC information for a particular `setQuality`. Each `qualityMetric` has an `accession` and the corresponding `name` are defined by the QC CV (see `qc-cv.obo`) and should be represented exactly as stated in the .obo file. The `value` carries the actual information.
+Metric `value`s can be either single values, tuple of values, or matrices or tables (shown in other examples). 
+
+Here we show two metrics: the first describes the relative total intensity of all contaminant proteins (from all runs in the set) -- the higher the value the more contaminants are present in the samples of the set.
+```
+            "accession": "QC:0000000",
+            "name": "Protein contaminant intensity ratio",
+            "value": 0.25
+```
+... and the second metric is a single column of a PCA result matrix. This metric is a bit special because it only partially describes the data.
```
bittremieux, line 103: I find this a weird way to encode PCA values. Wouldn't a table be a more natural representation? That way all PCA values are stored together.

doc/individual-runs.mzQC.md
cbielow: good catch. fixed

doc/set-of-runs.mzQC.md
cbielow: I totally agree!
But there is no way to encode the groups in a natural way that I can think of. There is another solution, but its also a bit complicated:
use a setQuality which contains all Raw files. In there, store the matrix as you suggest, and then use another qualityMetric to define the subgroups and their names.

bittremieux: Sorry I missed the call this morning. Stupid mistake. 🤦‍♂️ So I'm not sure to what extent this has been discussed, but at first thought I would prefer your second solution over the current one.

cbielow:
to go ahead on this: apart from how to represent supersets, this PR should add value.
So in order to reduce the number of things dangling in parallel, lets merge.
I will make suggestions for alternative solutions to the superset problem in another PR.

cbielow:
I updated the example now. It should reflect our agreement on how to represent groups.

(also fixed some invalid JSON example files)

bittremieux:
As @cbielow suggested, let's just merge this PR rather than leave it open indefinitely. @cbielow: Can you rebase against the current version of `main` and then just go ahead and merge it.

We can revisit all example files in the (near) future.

cbielow:
this should do it.
Can someone please confirm I did not miss any CV term numbers which needed updating? :)

Issue #119, opened 2021-03-24
Improved browsing of qc-cv.obo


sneumann:
Related to #107, I suggest to have a better browsing of `qc-cv`, 
which would also help #115 . I vaguely recall that the OpenMS team 
had some tools in OpenMS to create static HTML from OBO files. 
Resulting HTML could be hosted on github pages then.
Yours, Steffen

nilshoffmann:
Please see #122

nilshoffmann:
Documentation is now in my repositories CI, deployed to https://nilshoffmann.github.io/obo-adoc/qc-cv.adoc.html

sneumann:
The formatting looks good, but is lacking the navigationability. 
I found what I remembered:
http://htmlpreview.github.io/?https://github.com/nmrML/nmrML/blob/master/docs/mapping_and_cv.html
which is created by the OpenMS CVinspector in https://github.com/nmrML/nmrML/blob/master/Makefile#L127, unfortunately it requires a mapping file to work. A dummy mapping that allows the root term would work for the output. Yours, Steffen

nilshoffmann:
@sneumann @mwalzer 

So, should we keep 
1. The generated html for the complete CV using my script
2. The generated *mapped* html for the CV using the mapping file using the OpenMS tools?

Or should we just use the OpenMS tool for both?

bittremieux:
This might also be a relevant tool: http://vowl.visualdataweb.org/webvowl.html

We'd need to convert the OBO to OWL format, but those are mostly interchangeable. And the resulting interactive view looks pretty decent.

Ideally we should set something up with GitHub Actions to automatically generate an interactive webpage on all CV updates.

mwalzer:
would #122 also work after #141? PSI-MS-CV might appreciate a browse-friendly CV, too

nilshoffmann:
PSI-MS obo was one of my first use-cases for this, so yes, it should work. You just need to replace (or add) a line with the url for the psi-ms obo file. Here is a version from some time ago: https://nilshoffmann.github.io/obo-adoc/psi-ms.adoc.html

bittremieux:
As we are merging into PSI-MS-CV, it will be possible to easily navigate the QC terms in their own branch using OLS (and other tools that already index the PSI-MS-CV).

PR #122, opened 2021-03-25
Cross-linked Markup version of obo.

nilshoffmann:
If needed, I can provide the script to build this automatically as part of CI.


mwalzer:
I love it, perfect for bridging the time until we are ready for OLS integration. I need to go over the CI issues soon, anyway, so an automated update of the obo's representation would be bliss.

nilshoffmann:
Scripts for that are here: https://github.com/nilshoffmann/obo-adoc
Just need groovy (e.g. via sdkman: https://sdkman.io/install), edit the ontology-urls.ini (or recreate it containing just the link to https://raw.githubusercontent.com/HUPO-PSI/mzQC/master/cv/qc-cv.obo). Then run ./dtr.sh and copy the final target/qc-cv.adoc and target/qc-cv.adoc to your desired location.

nilshoffmann:
See #119 for an update on the current location of the qc-cv html.

bittremieux:
@nilshoffmann Could you try to update this PR to include:
- A GitHub Actions workflow so that the HTML page is automatically updated upon each commit/merge.
- Rather than including the raw HTML in the `cv` folder, make it so that the updated pages are hosted on our GitHub Pages website.

nilshoffmann:
The latest PR includes the required dependencies in the .build-tools folder and builds the obo html in the temporary build/target folder. The workflow also adds a Jekyll step which probably needs to be adjusted for the current jekyll page, which exists on the gh-pages branch.

mwalzer:
This is awesome @nilshoffmann . If I read that last step of the workflow right, it will deposit a `qc-cv.html` into the root of the gh-pages branch? Then, integration is simple with just adding that file to the _config.yml header_pages listing to make it a 1st class citizen. I don't have much experience with jekyll generally (this is partly why it took me so long to get the validator online), maybe you could give it a stab after this PR is merged?

nilshoffmann:
@mwalzer @bittremieux If we integrate with PSI-MS, do we still need the browseable html version?

bittremieux:
I agree, that's something to discuss. If the CV is on OBOFoundry, through integration with the MS CV, it's easily browsable on there. So that might solve that issue.

mwalzer:
Agreed, I don't think we need that. Yet, I think it's a nice feature, maybe at some point for the merged MS .obo?

bittremieux:
No longer necessary now we are [ready to merge into PSI-MS-CV](https://github.com/HUPO-PSI/psi-ms-CV/pull/99). Thanks for the effort @nilshoffmann!

PR #128, opened 2021-05-05
Add userParameter

### LINKED ISSUE #126: Add userParam
Provide a `userParam` type for temporary metrics that don't have a CV term (yet).

bittremieux:
Add a `userParameter` to allow metrics without a CV term:

- A new `userParameter` element for which only `name` is required (`accession` is optional, in contrast to `cvParameter`).
- A new `baseParameter` element from which both `userParameter` and `cvParameter` are derived, as they only differ based on whether or not `accession` is required.
- Replace all locations where `cvParameter` could occur by `baseParameter` to allow both `cvParameter` and `userParameter`.

Fixes #126.

**TODO**:

- [ ] Update the specification document accordingly.

### DIFF
--- a/schema/mzqc_schema.json
+++ b/schema/mzqc_schema.json
@@ -201,7 +201,7 @@
                 },
                 "fileFormat": {
                     "description": "Type of input file.",
-                    "$ref": "#/definitions/cvParameter"
+                    "$ref": "#/definitions/baseParameter"
                 },
                 "fileProperties": {
                     "description": "Detailed properties of the input file.",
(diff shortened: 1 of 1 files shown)


mwalzer:
I think we should also bump the version number. Since it is a bit more change than just minor changes in naming or spelling, maybe v1.1.0 ?

bittremieux:
Imo we should release a version 1.0.0 along with the publication of the specification document, not a version 1.x or something. We can consider the current version as a draft 1.0.0 that will be finalized with the public release of the specification document.

bittremieux:
[We will not add a `userParameter`](https://github.com/HUPO-PSI/mzQC/issues/126#issuecomment-843877399).

PR #131, opened 2021-06-14
Added example mzML file to support specdoc paragraph

mwalzer:
needs QC CV entries for `mzQC format` and `Min peaks for 50% of total spectra intensity` first

### DIFF
--- a/doc/mzml-mzqc-example.mzML
+++ b/doc/mzml-mzqc-example.mzML
@@ -1,0 +1,60 @@
+## Example of mzQC metrics in mzML
+
+```
+			<sourceFile id="QC1" name="BSA1_F1.mzQC" location="file:///">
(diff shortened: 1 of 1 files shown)


bittremieux:
Let's try to finalize this example and get it merged. @mwalzer can you work with @julianu to get the required CV terms in the OBO?

Maybe a consideration would be to not include the full mzML file in the repository to avoid bloat and just have the relevant snippet in the examples. I don't see that much added benefit from having the entire mzML file. If you do want to add it, @mwalzer please add it with git-lfs rather than git to keep the version tree a bit leaner (and overwrite the history in this branch to scrub that file).

mwalzer:
@bittremieux I picked a quite small mzML but I definitively see the point that the added value to Mb charged to our quota ratio is still poor. I will create a new PR w/ an excerpt and a compressed mzML and delete this branch.

Issue #133, opened 2021-06-14
[CV request] Min peaks for 50% of total spectra intensity 


mwalzer:
**Purpose:** A very basic metric assessing the quality of an individual spectrum by its peaks intensities

**Name:** Min peaks for 50% of total spectra intensity 

**Definition:** Minimum number of peaks to explain at least 50% of total spectrum intensity

**Comment:** None so far

**Value type:** `single value`

**Unit:** count

**Category:** ID free, MS metric,

mwalzer:
need it here #131

mwalzer:
Now in [PSI-MS](https://github.com/HUPO-PSI/psi-ms-CV/pull/107)

PR #138, opened 2021-07-21
added metric requested in #133

mwalzer:
Intended for #131 and its successor branch (TBD).

### DIFF
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -1,6 +1,6 @@
 format-version: 1.2
 data-version: 0.1.2
-date: 14:04:2021 12:39
+date: 21:07:2021 20:26
 saved-by: julianu
 default-namespace: QC
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
@@ -2095,6 +2095,18 @@ def: "The mean pump pressure in bar for the whole run." [PSI:QC]
 comment: Detecting inconsistencies in pump pressure between runs early helps troubleshooting and prevents instrument damage.
 is_a: QC:4000003 ! single value
 is_a: QC:4000010 ! ID free
+relationship: has_relation QC:4000013 ! QC metric relation: single run
+relationship: has_relation MS:1003019 ! pressure chromatogram
+relationship: has_relation QC:4000079 ! Pump pressure chromatogram
+
+[Term]
+id: QC:4000264
+name: Min. peaks for 50% of total spectra intensity
+def: "The minimum number of peaks to explain at least 50% of total spectrum intensity." [PSI:QC]
+comment: A very basic metric assessing the quality of an individual spectrum by its peaks intensities compared to the total spectrum intensity.
+is_a: QC:4000003 ! single value
+is_a: QC:4000010 ! ID free
+is_a: QC:4000015 ! QC metric relation: one spectrum
 relationship: has_relation QC:4000013 ! QC metric relation: single run
 relationship: has_relation MS:1003019 ! pressure chromatogram
 relationship: has_relation QC:4000079 ! Pump pressure chromatogram


cv/qc-cv.obo
```diff
@@ -2137,3 +2136,12 @@ property_value: has_optional_column: QC:4000249 ! PCA Dimension 3
 property_value: has_optional_column: QC:4000250 ! PCA Dimension 4
 property_value: has_optional_column: QC:4000251 ! PCA Dimension 5
 
+
+[Term]
+id: QC:4000268
+name: Min. peaks for 50% of total spectra intensity
```
bittremieux, line 2142: Maybe use the common term "total ion current" instead of "spectra/spectrum intensity"?

bittremieux: I was a bit quick with merging #139, so if we decide to change the term name, we shouldn't forget to modify the example mzML as well.

mwalzer: > we shouldn't forget to modify the example mzML as well.

Certainly.
> common term "total ion current" instead of "spectra/spectrum intensity"?

I was afraid that this might not be taken in local context and cause confusion that way. I'm open to change it. What do @cbielow @julianu think?

cv/qc-cv.obo
```diff
@@ -1,6 +1,6 @@
 format-version: 1.2
 data-version: 0.1.2
-date: 14:04:2021 12:39
+date: 21:07:2021 20:26
```
cbielow, line 3: can we (from now on) also increase the data-version, just to make sure people can pick the latest version without cross-checking with git/github?!

cv/qc-cv.obo
cbielow: is this metric for a single spectrum (how would you reference the spectrum in a runQuality unless you are using a table?) or for a whole experiment (all spectra)?
This should be made clear in the description.

Depending on the above, TIC would make sense to me if its many spectra. If its only one spectrum, then I'm acutally not sure, since TIC always suggests a chromatogram (for me).

cbielow: also: `name: Min. *number of* peaks for ...`
?

cv/qc-cv.obo
bittremieux: Yes, good point!

cv/qc-cv.obo
bittremieux: This is for a single spectrum. I think the term TIC is still correct in that case; the typical chromatogram plots are basically the TICs of all spectra combined.

I agree that it should be clearly specified that this is for a single spectrum only. Maybe this is even important enough to be a new categorization using a `relationship`. So rather than `single run`, as it is now, this would become `single spectrum` (with the corresponding new CV term). That seems more correct to me.

mwalzer: This sounds like a good path to a solution, but still leaves
> (how would you reference the spectrum in a runQuality unless you are using a table?)

which is another good point. So initially we thought this one up to use in a mzML with mzQC example. There, each spectrum has a cvParamGroup to accommodate the metric. But of course we probably want to report these or other `single spectrum` metrics. 
How about a collective `single spectrum`s table metric instead of a extra table metric for each `single spectrum` metric? Call it `single spectrum metrics table`. And the table is a mandatory column of spectrum identifiers plus any `single spectrum` metric as optional column. However, it is quite conceivable to me that some single spectrum metrics are used for one particular subset of spectra and some for another subset. Since repeating a metric within a run/set is not allowed, we have to use the same metric for both subsets and following some blank spots in the table. If I remember correctly we haven't specified explicitly what to do with missing values table columns but I'd argue N/A for the case where no value was calculated and nan where the metric calculation went off the beaten path. What do you think?

bittremieux: Section 9.3 discusses how to represent NaN. We don't mention N/A or null values yet.

A single spectrum table could be a solution. We'd have to think how to properly implement this though. For the optional columns, all of the possible values will need to be explicitly specified according to the current spec I think. And every time a new single spectrum metric is defined, this needs to be added as a optional column as well. That's going to become a mess pretty quickly. Can we somehow use inheritance or something in the CV to keep it maintainable?
The same kind of goes for the spectrum identifier: how do we represent multiple options (USI or nativeID, as mentioned in the spec)? One and only one should probably be used, so both columns shouldn't be specified as mandatory nor optional. Instead maybe two different single spectrum table terms in the CV, one when using USIs and one when using nativeIDs?

cbielow: All good points.

For the missing/uncomputable values:
nan seems to be unsuitable because it implies a number was expected, thus does not work well for strings etc.

How about: 
NULL for missing values 
NA for failed computation
?

This would be directly readable for R. Same for python?

bittremieux: JSON has a built-in `null` value, so that can directly be used.

JSON and Python don't have a built-in NA value. Would just `null` work as well, without distinguishing between the source of the missingness?

bittremieux:
Based on our discussion today I've defined the metric as a table now, which is how it would be included in mzQC files.

However, the columns of the table are the newly-defined USI column type and a count column type. So with these definitions we don't have have a term to associate with individual spectra, e.g. in mzML files.

cv/qc-cv.obo
```diff
@@ -2137,3 +2137,22 @@ property_value: has_optional_column: QC:4000249 ! PCA Dimension 3
 property_value: has_optional_column: QC:4000250 ! PCA Dimension 4
 property_value: has_optional_column: QC:4000251 ! PCA Dimension 5
 
+
+[Term]
+id: QC:4000268
+name: Number of peaks for at least 50% total ion current per spectrum
```
cbielow, line 2143: is this the "**Minimal** number of peaks explaining at least 50% of the total ion current of a spectrum"

cv/qc-cv.obo
```diff
@@ -2137,3 +2137,22 @@ property_value: has_optional_column: QC:4000249 ! PCA Dimension 3
...
+is_a: QC:4000010 ! ID free
+is_a: QC:4000022 ! chromatogram metric
+relationship: has_relation QC:4000013 ! QC metric relation: single run
+property_value: has_column: QC:4000269 ! Universal Spectrum Identifier
+property_value: has_column: QC:4000115 ! Count
+
+[Term]
+id: QC:4000269
+name: Universal Spectrum Identifier
+def: "The column contains Universal Spectrum Identifiers to refer to individual spectra." [PSI:QC]
```
cbielow, line 2156: mention "USI" somewhere? To make it searchable. E.g. imagine someone who sees "This is an USI, used for..." and has no clue what it means...

mwalzer:
I think this PR is ready now. @bittremieux  @cbielow what do you think?

mwalzer:
After more discussion and some superseding changes to the CV organisation, this is now outdated (see https://github.com/HUPO-PSI/psi-ms-CV/issues/106)

bittremieux:
@mwalzer I think we also need to update this new term in the `mzml-mzqc-example.mzML` example, right?

mwalzer:
:frowning_face: Yes, I noticed this week that we actually have to update the examples quite a bit (at least in terms of `[Terms]`). That's why I opened two new request issues with terms we have been using in the examples and another one or two to follow soon.

PR #139, opened 2021-07-21
Added example mzML file to support specdoc paragraph

mwalzer:
This is the successor PR to #131 with #133 considered.
@bittremieux  : I clipped the mzML to 3 spectra, does not bloat, gets the point across. And it's valid 1.1.0 mzML! :partying_face:


doc/examples/mzml-mzqc-example.mzML
```diff
@@ -0,0 +1,448 @@
...
+		<fileContent>
+			<cvParam cvRef="MS" accession="MS:1000294" name="mass spectrum" />
+		</fileContent>
+		<sourceFileList count="1">
+			<sourceFile id="sf_ru_0" name="BSA1_F1.mzML" location="file:///home/walzer/ms-tools/openms-source-git/share/OpenMS/examples/FRACTIONS">
+				<cvParam cvRef="MS" accession="MS:1000569" name="SHA-1" value="ce02b0744755a8d026f11ece3c5f9b76efcd87f8" />
+				<cvParam cvRef="MS" accession="MS:1000584" name="mzML format" />
+				<cvParam cvRef="MS" accession="MS:1000777" name="spectrum identifier nativeID format" />
+			</sourceFile>
+			<sourceFile id="QC1" name="BSA1_F1.mzQC" location="file:///">
```
bittremieux, line 21: Is this file location expected to be empty?

doc/examples/mzml-mzqc-example.mzML
```diff
@@ -0,0 +1,448 @@
...
+		</software>
+		<software id="so_dp_sp_1_pm_2" version="1.6.0" >
+			<cvParam cvRef="MS" accession="MS:1000758" name="FileMerger" />
+		</software>
+		<software id="so_dp_sp_1_pm_3" version="1.6.0" >
+			<cvParam cvRef="MS" accession="MS:1000757" name="FileFilter" />
+		</software>
+		<software id="so_dp_sp_1_pm_4" version="2.1.0" >
+			<cvParam cvRef="MS" accession="MS:1000757" name="FileFilter" />
+		</software>
```
bittremieux, line 72: Should the QC generating software be included here?

bittremieux (commented):
Two minor comments. Looks good!

doc/examples/mzml-mzqc-example.mzML
mwalzer: Was in the root of the container I built for creating the example. Did not include the filename, just the base path. The spec doc says about the location attribute:
> URI-formatted location where the file was retrieved.

I can see how this can be interpreted both base path and full path.
Maybe for our spec doc definition of locations, we should be more explicit? Right now we say for name and location:
>The name MUST uniquely match to a location (specified below) listed in the mzQC file.

And

>Unique file location, REQUIRED to be specified as a URI. The file URI is RECOMMENDED to be publicly accessible.
Should we require name 'redundancy'?

mwalzer: I changed it to read `file:///examples/`

bittremieux: Yes, my reading of the spec was always that `location` was supposed to be partially redundant with `name`. I've clarified it a bit in our spec name, check it out.

bittremieux (approved):
Looks good with those final changes, thanks!

PR #144, opened 2021-09-07
Fixed typo for QC:4000259

poshul:
Fix typo of is_a: QC:4000023


cv/qc-cv.obo
```diff
@@ -2038,7 +2038,7 @@ name: MS1 density per quantile
 def: "The first to n-th quantile of MS2 scan peak counts." [PSI:QC]
```
bittremieux, line 2038: Does this then also need to be "MS1" here?

bittremieux (approved):

Issue #141, opened 2021-08-11
Merge CV into PSI-MS CV


bittremieux:
As per feedback from the PSI internal review, the CV should be merged with the [PSI-MS CV](https://github.com/HUPO-PSI/psi-ms-CV) prior to external review.

mwalzer:
Following up on the call with PSI-MS-Vocab group:
- [ ]  We switch to MS prefix but keep the accession number space.
- [ ]  We attach our root element to the top common PSI-MS CV element
- [ ]  We switch from property_value: to relationship:
- [ ]  We double check for 'duplicates' like retention time, and amend the name with 'column' where applicable or change the name accordingly (I don't think there are any clashes that can't be fixed with ... Column.)
- [ ]  We use PSI-MS has_unit instead of our has_units

bittremieux:
A [PR with the first terms to be merged into the PSI-MS-CV](https://github.com/HUPO-PSI/psi-ms-CV/pull/99) has now been opened. Hoorah! 🎉

Issue #120, opened 2021-03-24
Add more intermediate level classification of metrics


sneumann:
Related to #115, I would like to see a somewhat deeper nesting of metrics,
e.g. a category on PPM (or mass error) based ones. Or I might not 
have found them in the OBO. Yours, Steffen

bittremieux:
We have now extended and harmonized the QC terms categorization for the first batch of terms that are going to be merged into the PSI-MS-CV (#165). Future term merging will be categorized similarly.

PR #145, opened 2021-10-12
preparations for cv merge to psi-ms

mwalzer:
see #141

### DIFF
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -14,7 +14,7 @@ import: http://ontologies.berkeleybop.org/uo.obo
 import: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
 import: https://raw.githubusercontent.com/obi-ontology/obi/master/obi.owl
 import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato.obo
-ontology: qc
+ontology: ms
 
 [Typedef]
 id: has_relation
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -635,7 +635,8 @@ relationship: has_units UO:0000189 ! count unit
 [Term]
 id: QC:4000116
 name: Peptide sequence column
-def: "The column contains peptide sequences; multiple peptides are separated by space." [PSI:QC]
+def: "The column contains peptide sequences; multiple peptides are separated by space; sequences in ProForma encoding." [PSI:QC]
+comment: ProForma v2 - https://github.com/HUPO-PSI/ProForma/
 is_a: QC:4000107 ! Table column type
 relationship: has_type NCIT:C45253 ! String
 
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -43,8 +43,8 @@ name: has_units
 
 [Term]
 id: MS:4000000
-name: PSI-MS CV Quality Control Vocabulary
-def: "PSI-MS CV Quality Control Vocabulary." [PSI:MS]
+name: QC CV term
+def: "PSI Quality Control controlled vocabulary term." [PSI:QC]
 relationship: part_of MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
 
 [Term]
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -576,8 +576,298 @@ is_a: MS:4000022 ! chromatogram metric
 relationship: has_relation MS:4000014 ! QC metric relation: multiple runs
 
 [Term]
+id: QC:4000082
+name: isolation window target m/z
+def: "The primary or reference m/z about which the isolation window is defined." [PSI:QC]
+comment: The isolation window target m/z is read straight from the mzML file and reported for every unique isolation window. This serves mainly as information to orientate the user.
+xref: value-type:xsd\:double "The allowed value-type for this CV term."
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000010 ! ID free
+is_a: QC:4000024 ! MS2 metric
+relationship: has_relation MS:1000827 ! isolation window target m/z
+
+[Term]
+id: QC:4000083
+name: scans per isolation window
+def: "The number of scans collected during the run reported for all unique isolation windows." [PSI:QC]
+comment: This is mainly informative and serves as a nice confirmation that mzML contains all the necessary information.
+xref: value-type:xsd\:int "The allowed value-type for this CV term."
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000010 ! ID free
+is_a: QC:4000024 ! MS2 metric
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+
+[Term]
+id: QC:4000084
+name: average m/z range of isolation window
+def: "The average range (isolation window upper limit - isolation window lower limit) in m/z for each unique isolation window." [PSI:QC]
+comment: This is the range set by the user in the isolation scheme. The purpose is to provide context for other metrics.
+xref: value-type:xsd\:double "The allowed value-type for this CV term."
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000010 ! ID free
+is_a: QC:4000024 ! MS2 metric
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+
+[Term]
+id: QC:4000085
+name: proportion of total TIC per isolation window
+def: "The TIC for all occurrences of the relevant isolation window in the run is summed and divided by the total TIC for all MS2 scans in the run." [PSI:QC]
+comment: A larger value indicates that isolation windows with that particular target m/z contributed more to the total TIC.
+xref: value-type:xsd\:double "The allowed value-type for this CV term."
+is_a: QC:4000004 ! n-tuple
+is_a: QC:4000010 ! ID free
+is_a: QC:4000024 ! MS2 metric
+relationship: has_relation QC:4000013 ! QC metric relation: one run
+relationship: has_relation QC:4000070 ! MS2 Total ion current chromatogram
+relationship: has_relation QC:4000082 ! isolation window target m/z
+
+[Term]
+id: QC:4000086
+name: average isolation window density
+def: "For each unique isolation window this value reports the number of ions detected on average throughout the run." [PSI:QC]
+comment: A higher value indicates that a higher number of ions was detected for this specific isolation window on average.
+xref: value-type:xsd\:double "The allowed value-type for this CV term."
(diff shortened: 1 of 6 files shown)


cv/qc-cv.obo
```diff
@@ -42,336 +40,330 @@ name: has_tuple
 id: has_units
 name: has_units
 
+
+[Term]
+id: MS:4000000
+name: PSI-MS CV Quality Control Vocabulary
+def: "PSI-MS CV Quality Control Vocabulary." [PSI:MS]
+is_a: MS:0000000 ! Proteomics Standards Initiative Mass Spectrometry Vocabularies
```
timosachsenberg, line 48: minor observation: so far never used with is_a
https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L9565

mwalzer: Thanks, good point. We'll have to fix that.

mwalzer:
@julianu @bittremieux @cbielow Please let me know if you think we can use the PR as it is for a merge request with PSI-MS-CV or not and where you think we need additional structure or better names/definitions.

nilshoffmann:
@mwalzer I prepared pull request #159 against this one which resolves some remaining QC prefixes and removes and extra id

sneumann:
Hi, I am doing some stupid things here: 
```
cut -d : -f 1 qc-cv.obo  | sort | uniq -c | sort -n 
...
    234 [Term]
    237 def
    246 id
    246 name
    602 is_a
```
and trying to see why we have more `Term`s than `def`s . Found a few issues:

1) id: MS:4000107 has two [Term]s
1) QC:4000027 and QC:4000028 have no def
1) five trailing spaces (don't think they really matter)
```
grep -n '^\[Term\] $' qc-cv.obo # line numbers
2193:[Term] 
2265:[Term] 
2281:[Term] 
2297:[Term] 
2303:[Term] 
```
I see several CV issues https://github.com/HUPO-PSI/mzQC/issues?q=is%3Aissue+is%3Aopen+label%3ACV but several of them are adressed in commits by Matthias. Can they be closed ? 
I focused on such technical things, little content review done. 
Yours, Steffen

bittremieux:
Great, thanks for checking those issues @sneumann. I think I've fixed them (with a small git snafu in the mix).

Most of the open CV issues can indeed be closed when this branch is merged in.

bittremieux (commented):
Some comments/questions:

- [ ] Should the namespace coverage be `sequence(7,4000000,4999999)` instead?
- [ ] `has_relation` is a weird name for the typedef. It doesn't exist in PSI-MS-CV yet. Can we bring it in line with that CV or rename it to something better?
- [ ] `has_type` should be renamed to `has_value_type` for consistency with PSI-MS-CV.
- [ ] `has_tuple` is not used and can be removed.
- [ ] "ID free" etc. is a metric category imo, and should be specified using `relationship` rather than `is_a`.
- [ ] A bunch of metrics are missing unit and type specifications that could be included.
- [ ] `MS:4000055` and `MS:4000056`: The name uses "quantile", but the definition details that it's quartiles.

The CV is too unwieldy to fully go through it. At least I have not been able to do so.
Unfortunately I think that the CV would benefit from carefully reviewing all QC terms individually with a checklist in hand, because several terms that I looked at so far still have issues:
- Is the metric properly defined?
- Have all potential relationships and categorizations been specified?
- Has the unit been specified?
- Has the value type been specified?
- Have links to other terms been listed?
- Has the term been checked for duplication in other CVs?

sneumann:
Regarding "Is the metric properly defined?" @tnaake could comment on the ones he found most difficult to understand while implementing them. Yours, Steffen

tnaake:
Hi all and many thanks for the fantastic conceptualization of MS QC metrics,

I am currently implementing together with @jorainer parts of the mzQC metrics in `R` (see https://github.com/tnaake/MsQuality for the repo). 

We are currently focusing on the metrics that can be calculated by the usage of `Spectra` (https://github.com/rformassspectrometry/Spectra) and `MsExperiment` (https://github.com/rformassspectrometry/MsExperiment) objects as these are the central architecture to store MS data in R. 

Looking at some of the metrics, at least I was a bit puzzled how to calculate some of them. It would be great if you could comment if we (I) interpreted those correctly (and possibly adjust the definitions for easier understanding). Please note, that for some of the metrics we currently deviate from the definitions. Some definitions e.g. leave it open if a certain quantile needs to be calculated, while we focus for some functions on the `0`, `.25`, .5`, `.75`. 1` quantile. Please also note, that we are dealing here sometimes explicitly with `NA` values. When dealing with real-world data sets, we expect that there are no `NA`s. 

 - RT over TIC quantile (QC:4000054): 
     -  (1) the `Spectra` object is ordered according to the retention time,
     -  (2) the cumulative sum of the ion count is calculated (TIC),
     -  (3) the quantiles are calculated according to the given quantiles, e.g. the 0%, 25%, 50%, 75% and 100% quantile,
    - (4) the retention time/relative retention time (retention time divided by the total run time taking into account the minimum retention time) is calculated,
    -  (5) the (relative) duration of the LC run after which the cumulative TIC exceeds (for the first time) the respective quantile of the cumulative TIC is calculated and returned.

- MS1/MS2 quantiles RT fraction (QC:4000055/QC:4000056):
    - (1) the retention time duration of the whole `Spectra` object is determined (taking into account all the MS levels),
    - (2) the `Spectra` object is filtered according to the MS level and subsequently ordered according to the retention time
    - (3) the MS events are split into four (approximately) equal parts,
    - (4) the relative retention time is calculated (using the retention time duration from (1) and taking into account the minimum retention time),
    - (5) the relative retention time values associated to the MS event parts are returned.

- MS1 quantile TIC change ratio to Quantile 1 (QC:4000057) or to previous quantile (QC:4000058):
    - *TIC changes* are interpreted as follows:
        - (1) the cumulative sum (`cumsum`) of the  `spectra`'s TIC is calculated  (with spectra ordered by retention time),
        - (2) quartiles are then calculated on these, 
        - (3) for *QC:4000057* the log2 ratio between the 25, 50, 75 and 100% quantile to the 0% quantile is calculated. For *QC:4000058*  ratios between the 25/0, 50/25, 75/50 and 100/75% quantiles are calculated (this function interprets the *quantiles* from the [PSI:QC] definition as *quartiles*, i.e. the 0, 25, 50, 75 and 100% quantiles are used),
        - The `log2` values are returned instead of the `log` values.

- Peptide identification rate of the interquartile RT period (QC:4000073):
    - (1) the `Spectra` object is filtered according to the MS level,
    - (2) the retention time values are obtained,
    - (3) the 25% and 75% quantiles are obtained from the retention time values (`NA` values are removed),
    - (4) the number of eluted features between this 25% and 75% quantile is calculated,
    - (5) the number of features is divided by the interquartile range of the retention time and returned.

- Area under TIC RT quantiles (QC:4000078):
    - (1) the `Spectra` object is filtered according to the MS level, 
    - (2) the `Spectra` object is ordered according to the retention time, 
    - (3) the 0%, 25%, 50%, 75%, and 100% quantiles of the retention time values are obtained, 
    - (4) the ion count of the intervals between the 0%/25%, 25%/50%, 50%/75%, and 75%/100% are obtained (this function interprets the *quantiles* from the [PSI:QC] definition as *quartiles*, i.e. the 0, 25, 50, 75 and 100% quantiles are used),
    - (5) the ion counts of the intervals are summed (TIC) and the values returned.

- Median of TIC values in the RT range in which the middle half of peptides are identified (QC:4000130):
    - (1) the `Spectra` object is filtered according to the MS level, 
    - (2) the `Spectra` object is ordered according to the retention time,
    - (3) the features between the 1st and 3rd quartile are obtained (half of the features that are present in `spectra`),
    - (4) the ion count of the features within the 1st and 3rd quartile is obtained,
    - (5) the median value of the ion count is calculated (`NA` values are removed) and the median value is returned.

- Median of TIC values in the shortest RT range in which half of the peptides are identified (QC:4000132):
    - (1) the `Spectra` object is filtered according to the MS level,
    - (2) the `Spectra` object is ordered according to the retention time, 
    - (3) the number of features in `spectra` is obtained and the number for half of the features is calculated,
    - (4) iterate through the features (always by taking the neighbouring half of features) and calculate the retention time range of the set of features,
    - (5) retrieve the set of features with the minimum retention time range,
    - (6) calculate from the set of (5) the median TIC (`NA` values are removed) and return it.

Many thanks, 
@tnaake

bittremieux:
Very nice, I'm excited to see this work!

I interpret the metrics the same way as you do, although I have not implemented any of these myself.

The reference implementation for the first three metrics is QuaMeter IDfree. QuaMeter IDfree is straightforward to get working, so you should be able to compare your results against theirs relatively easily.

The final four metrics are from NIST MSQC I think. I would not recommend trying to get NIST MSQC working, if that even still exists. But these metrics are also implemented in QuaMeter IDbased mode. It takes a bit more effort to get those results because you need to have identification data as well, so ymmv with comparing against that.

Pinging @dtabb73, who has supervised development of QuaMeter.

tnaake:
Dear @bittremieux 

many thanks for checking your interpretation of the metrics with ours. We also hope that the package will be used within the R community heavily as it represents an interface with the `Spectra` and `MsExperiment` and the concepts of mzQC and are looking forward to getting some end user feedback after releasing the package. 

We will have a look at QuaMeter IDfree and will check if we get the same results. Thanks for the hint!

Best wishes,
@tnaake

mwalzer:
Hi all,
as @bittremieux said: very unwieldy. I think to help progress, we'll merge this one to get the main branch updated with the progress made in this PR, then move in smaller steps (and more focused PRs). I'd suggest we prepare thematic stripes of terms to go into PSI-MS first to keep the size manageable, using the checklist above. 

@tnaake @sneumann Also very excited about your work! I'll `@ mention` in the future PRs that relate to your comments if you don't mind.

mwalzer:
And just to complete a follow-up:
> Should the namespace coverage be sequence(7,4000000,4999999) instead?

Doesn't really matter once we're in PSI-MS where we don't have prefix exclusivity

> has_relation is a weird name for the typedef. It doesn't exist in PSI-MS-CV yet. Can we bring it in line with that CV or 
rename it to something better?

we should find a solution in the telco.

> has_type should be renamed to has_value_type for consistency with PSI-MS-CV.

Yep, I vaguely remember the last telco we agreed to that.

> has_tuple is not used and can be removed.

:+1:

> "ID free" etc. is a metric category imo, and should be specified using relationship rather than is_a.

We should find a solution in the telco. I think another agenda point would be the first bunch of terms. Maybe by category? Like that we can also introduce more structure as we go along. (#120) 

> MS:4000055 and MS:4000056: The name uses "quantile", but the definition details that it's quartiles.

See, that's the kind of review we need. We need to catch those. See PR #161

> A bunch of metrics are missing unit and type specifications that could be included.

A list would be great.

PR #161, opened 2021-12-08
fixing MS:4000055 and MS:4000056

mwalzer:
1. quantiles -> quarters (that's what is used, really, not the division points but the whole section, in this case, quarters)
2. also made the definition more specific using "The **time** interval ...".

### DIFF
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -259,8 +259,8 @@ relationship: has_units UO:0000191 ! fraction
 
 [Term]
 id: MS:4000055
-name: MS1 quantiles RT fraction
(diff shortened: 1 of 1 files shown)


bittremieux:
Should the correct term be "quartile" instead of "quarter"?

bittremieux:
@mwalzer Can we quickly look at the changes in this PR? These terms are included in the [PSI-MS-CV merging PR](https://github.com/HUPO-PSI/psi-ms-CV/pull/99) with their old definitions. If necessary, the changes proposed here can still quickly be applied. These are just minor changes, but they make sense imo. Although I'd still use "quartile" instead of "quarter" in the names and definitions.

Issue #83, opened 2019-09-24
Introducing comments to the CV entries


julianu:
To better understand the interpretation of the CV entries, or rather the values in the mzQC file itself, I would like to introduce comments to the obo files.
This will be in the next pull request (containing the entries from issue #52 ) and needs to be done for all metrics. Help is very welcome for this.

mwalzer:
:+1: 
So I guess this will be a thing of ongoing refinement: so should we suggest comments/changes in the respective original cv issue, new issue, PR or elsewhere?
Interesting though is to have an open thread for each cv term. Worry with this is it might get too long and confusing.

mwalzer:
wiki might be another place, but it's bad for discussions.

mwalzer:
This might actually help us some 
1. present our terms as clear as possible for a #141 PR
2. our terms being mixed in with general MS terms, PEFF terms, etc. (after #141) it will help avoid misinterpretation in the future even more

gets a priority :+1: from me

bittremieux:
[As discussed recently](https://github.com/HUPO-PSI/mzQC/blob/main/meeting_notes/2022/20220112_telco.md), we don't want to overdo it and introduce unnecessary comments. However, we do want to include comments when they can help to clarify how to interpret metrics.

The first batch of terms proposed for inclusion into the PSI-MS-CV illustrates this perfectly. Although this remains an ongoing point of attention as we merge more terms into the PSI-MS-CV, I'll close this issue for now.

PR #165, opened 2021-12-10
packaged ID-free metrics from QuaMeter for a first PR into MS-CV

mwalzer:
We now have a more tree-like hierarchy, as 'somewhat similar to is_a' tags are relegated to relationships. There, `has_relationship` is mostly pruned away or converted into `is_a` where there is good inheritance indication. We discussed `associated_with`, sofar no occasion has arisen for that with the QuaMeter ID-free metrics yet. Overall, this results in a less conflated structure. It also means that 'metric discovery' will work less well by DAG traversal (`is_a` builds that) and more via search terms of relations (which IMO is the better way, but doesn't visualise so nicely). As another outcome of this PR I think it would be nice to have derive some CV policies.

About other stanza tags used:
As discussed in the last telcos, we needed more concise names for the relations we want to describe for the terms. The `has_metric_category` and `has_value_concept` give important clues to plotting in a generic setting and unsupervised useage/search. Simple but instructive `has_value_type` will only inform on the data-type, though ultimately this is each implementers own choice. Other info on the values' properties is left to `has_value_concept`. For example informing that the reported values are quantiles. To be complete units should also be provided, `has_columns` will provide value_type, unit and value_concept for tables per  column. In short, here the checklist again:
    Is the metric properly defined?
    Have the is_a specifications proper predecessors and QC metric value types? 
    Has the unit been specified?
    Has the value type been specified?
    Has the value category been specified?
	has_relation -> is_a/remove as non-essential/associated_with

Other notes:
Back when I drafted the Quameter metric representations in our CV, I took the QuameterManual listing of ID-free metrics. Some terms clearly have comments that are not intended to end up in the vocabulary, and indicate where we'd benefit in (definition) refinement from the metrics original creators (@DTabb). I am of the opinion that the comments are a good place to give valuable insight to metric use and special implementation remarks.

Alas, how we organise the additional information in relationships is not without ambiguity. Here are two things we should consider for the value definition of metrics in this PR (and following):
UO:0000191 ! fraction - unit or value_concept? Same for UO:0000189 ! count unit (not all counts though, see density). Same for 'frequencies', hertz implies a frequency. Which way to use and how to avoid mixup in the future?

### DIFF
--- a/cv/qc-cv.obo
+++ b/cv/qc-cv.obo
@@ -1,13 +1,14 @@
 format-version: 1.2
 data-version: 0.1.5
-date: 10:11:2021 14:03
+date: 09:12:2021 12:03
 saved-by: Mathias Walzer
 default-namespace: MS
 namespace-id-rule: * MS:$sequence(7,0,9999999)$
 remark: coverage of namespace-id: MS:$sequence(7,4000000,4999999)$: Mass spectrometry quality control metrics
 remark: creator: Chris Bielow <chris.bielow <-at-> fu-berlin.de>
+remark: creator: Wout Bittremieux <wbittremieux <-at-> health.ucsd.edu>
+remark: creator: Nils Hoffmann < nils.hoffmann <-at-> cebitec.uni-bielefeld.de>
 remark: creator: Julian Uszkoreit <julian.uszkoreit <-at-> ruhr-uni-bochum.de>
-remark: creator: Martin Eisenacher <martin.eisenacher <-at-> ruhr-uni-bochum.de>
 remark: creator: Mathias Walzer <walzer <-at-> ebi.ac.uk>
 remark: namespace: MS
 import: http://ontologies.berkeleybop.org/uo.obo
@@ -15,6 +16,7 @@ import: https://raw.githubusercontent.com/HUPO-PSI/psi-ms-CV/master/psi-ms.obo
 import: https://raw.githubusercontent.com/obi-ontology/obi/master/obi.owl
 import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato.obo
 ontology: ms
+
 
 [Typedef]
 id: has_relation
@@ -25,6 +27,14 @@ id: has_type
 name: has_type
 
 [Typedef]
+id: has_units
+name: has_units
+
+[Typedef]
+id: has_value_concept
+name: has_value_concept
+
+[Typedef]
 id: has_column
 name: has_column
 
@@ -32,9 +42,6 @@ name: has_column
 id: has_optional_column
 name: has_optional_column
 
-[Typedef]
-id: has_units
-name: has_units
 
 
 [Term]
@@ -105,12 +112,6 @@ id: MS:4000011
 name: quantification based
 def: "QC metric based on quantification results." [PSI:QC]
 is_a: MS:4000008 ! QC metric category
-
-[Term]
-id: MS:4000012
-name: QC metric relation DEPRECATED
-def: "The basis for the metric calculation, such as 'single run' or 'single spectrum'." [PSI:QC]
-relationship: part_of MS:4000000 ! PSI-MS CV Quality Control Vocabulary
 
 [Term]
 id: MS:4000013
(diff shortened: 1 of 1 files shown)


mwalzer:
For your convenience the stanza tag order (see full in obo1.4 specification):
    id
    name
    def
    comment
    synonym
    is_a
    relationship

mwalzer:
Fraction, I think is in UO because it describes when a ratio of two values of the same unit is formed, result basically being unitless. Hence, e.g. if a a term is a (value_concept) log ratio, fraction as unit might be beneficial some times to, say, find similar ones? As for log ratio, it is like with fraction but the other way around. But I think it is mostly a concept. And in regards to metric discovery/unsupervised use, it might be good to have **both** in a relationship tag (i.e. add some duplicate info). A pot. example is MS:4000054 ! RT over TIC quantile. 
	Now how to boil this into a CV policy easy to follow in future and keep the QC terms consistant and easy to search?

cv/qc-cv.obo
```diff
@@ -24,6 +26,14 @@ name: has_relation
 id: has_type
 name: has_type
```
bittremieux: Typedef not used, remove?

cv/qc-cv.obo
```diff
@@ -32,9 +42,6 @@ name: has_column
 id: has_optional_column
 name: has_optional_column
 
-[Typedef]
-id: has_units
-name: has_units
 
 
 [Term]
```
bittremieux: Line 61 (GitHub doesn't let me add a comment there directly) should be renamed to "PSI-MS CV Quality Control Vocabulary" for consistency with PSI-MS-CV and the relationships later on here.

bittremieux: Line 68 and following: Instead of "OBI:0001442 ! q-value", a better (similar) example would be one of the q-value terms from the PSI-MS-CV. For example: MS:1002354 ! PSM-level q-value

bittremieux: Line 81 and following: Remove comments as per PSI-MS-CV advice?

cv/qc-cv.obo
```diff
@@ -106,12 +113,6 @@ name: quantification based
...
 
-[Term]
-id: MS:4000012
-name: QC metric relation DEPRECATED
-def: "The basis for the metric calculation, such as 'single run' or 'single spectrum'." [PSI:QC]
-relationship: part_of MS:4000000 ! PSI-MS CV Quality Control Vocabulary
-
 [Term]
 id: MS:4000013
 name: single run based
```
bittremieux: Line 149: "A metric" -> "QC metric" for consistency.

cv/qc-cv.obo
```diff
@@ -190,2207 +191,270 @@ name: MS metric
...
 synonym: "MS2-PrecZ-3" RELATED []
 synonym: "MS2-PrecZ-4" RELATED []
 synonym: "MS2-PrecZ-5" RELATED []
 synonym: "MS2-PrecZ-more" RELATED []
+is_a: MS:4000006 ! table
+relationship: has_metric_category MS:4000013 ! single run based
+relationship: has_metric_category MS:4000010 ! ID free
+relationship: has_metric_category MS:4000024 ! MS2 metric
+relationship: has_metric_category MS:4000025 ! ion source metric
+relationship: has_column: MS:4000238 ! Charge state column
```
bittremieux: Based on some earlier feedback from the PSI-MS-CV folks, they were reluctant to have to redefine terms used for columns as new terms, but rather suggested just using existing terms directly as columns. This prohibits full validation of column types, but doesn't have any major downsides as far as I understand? Maybe not a bad thing for simplicity of the CV and to avoid near-duplication? See Eric's comment in #164.

bittremieux (commented):
- [ ] Resolve comments. Possibly integrate them in the definition instead to get rid of them? See Eric's comment in [PSI-MS-CV#89](https://github.com/HUPO-PSI/psi-ms-CV/issues/89#issuecomment-983192909). See also #83.
- [x] Fraction is not a unit, so should not be specified as such. Semantically there's no real difference between a fraction and a (log) ratio, yet the former is specified with `has_units` and the latter with `has_value_concept`. Instead, both should probably be `has_value_concept`s?
- [x] Accession renumbering? Some of the accessions are out of order and gapped now, do we want to reorder them?

bittremieux:
[PR 99](https://github.com/HUPO-PSI/psi-ms-CV/pull/99) now opened in the PSI-MS-CV repository to merge these terms in there. Thank you @mwalzer.

PR #169, opened 2022-02-15
packaged ID-based metrics (and some ID-free) for a second PR into MS-CV

mwalzer:
Mostly counts, but also some contrasts of id-free and their id-based portions (e.g. median precursor charge for MS2 spectra / identified)


dtabb73:
I am taking a look at this pull request, and I'm not seeing any problems (except that I barely understand what a pull request represents).

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 id: MS:4000068
-name: Ambient humidity
-def: "The ambient relative humidity in percent (0,100) over one or more time points. The first column contains is the ambient humidity value(s). The second contains the respective retention time. A negative RT implies that the exact time of measurement is unknown." [PSI:QC]
-is_a: MS:4000006 ! table
-is_a: MS:4000010 ! ID free
-is_a: MS:4000026 ! environment metric
-relationship: has_relation MS:4000013 ! single run based
-relationship: has_column: MS:4000108 ! Retention time
-relationship: has_column: MS:4000260 ! Percent
+name: number of PSM
```
julianu: In consistancy with the next item, add "before FDR".
I would rather suggest even to call it something like "number of PSMs (unfiltered)" to make it more general (also in the following term for spectra)

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+def: "The number of peptide-to-spectrum-matches the identification process provides (before FDR filtering)." [PSI:QC]
+is_a: MS:4000003 ! single value
+relationship: has_metric_category MS:4000009 ! ID free metric
```
julianu: PSMs is probably a ID based metric.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-name: Fraction of peptide IDs around max MS1
-def: "Fraction of all peptides identified at least X minutes earlier or later than the most intense MS1 signal of the identification. The first column contains the time in minute-wise steps of seconds, the other comlumn the fraction of identified. Negative time values indicate minutes before and positive after maximum MS1." [PSI:QC]
-comment: Estimates very early peak broadening.
-is_a: MS:4000006 ! table
-is_a: MS:4000009 ! ID based
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:4000013 ! single run based
-relationship: has_column: MS:4000108 ! Retention time
-relationship: has_column: MS:4000239 ! Fraction
+name: number of identified peptides
```
julianu: To be consistand add "after FDR" to the name. The def describes it correctly.
Applies also to the following 5 terms.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_units UO:0000189 ! count unit
 
 [Term]
 id: MS:4000073
-name: Peptide identification rate of the interquartile RT period
-def: "The identification rate of peptides for the interquartile retention time period, in peptides per second." [PSI:QC]
-comment: Higher rates indicate efficient sampling and identification.
+name: number of proteins and groups from all PSM
+def: "Number of proteins identified from all PSM at given FDR threshold, groups count as one." [PSI:QC]
```
julianu: This is a bit unclear to me. Shoudl this just be "protein groups" also in the name and counting grouped accessions/proteins as just the group entity?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_units UO:0000189 ! count unit
 
 [Term]
 id: MS:4000074
-name: Median MS1 peak FWHM for peptides
-def: "Median of all MS1 peak widths at half maximum (FWHM) for all identified peptides, in seconds" [PSI:QC]
-comment: Sharper peak widths (i.e. smaller values) indicate better chromatographic resolution.
+name: number of distinct proteins from all PSM
+def: "Number of distinct proteins without duplicate count or groups from all PSM at given FDR threshold." [PSI:QC]
```
julianu: ... from teh comment before: and here  count all accessions in the respective protein groups?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_units UO:0000189 ! count unit
 
 [Term]
 id: MS:4000075
-name: Interquartile distance of MS1 peak FWHM for identifications
-def: "Interquartile distance of all MS1 peak widths at half maximum (FWHM) for all identifications, in seconds" [PSI:QC]
-comment: Tighter distributions indicate better consistency.
+name: number of undistinguishable proteins groups from all PSM
```
julianu: "protein groups" instead of "proteins groups" (and add "after FDR")

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 id: MS:4000076
-name: Median MS1 peak FWHM for peptides in RT quantile
-def: "Median of all MS1 peak widths at half maximum (FWHM) for all identified peptide, in seconds, in the respective identifications' quantiles ordered by retention time. The data is given in a tuple defining the quantiles, 4-tuple represents quartiles, 10-tuple deciles, etc." [PSI:QC]
-comment: Sharper peak widths (i.e. smaller values) indicate better chromatographic resolution in the given quantile.
-is_a: MS:4000004 ! n-tuple
-is_a: MS:4000009 ! ID based
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:1000086 ! full width at half-maximum
-relationship: has_relation MS:4000013 ! single run based
+name: ID ratio
```
julianu: maybe better "ratio of identified spectra"?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+def: "The coverage of protein sequences from the peptide sequences identified. The table records the coverage itself (in percent), the protein accession (SHOULD correspond to the accession used in fasta document used for Sequence DB based identification), length of the protein (, and optionally if it is 'target' or 'decoy' entry and how many PSM can be attributed to this protein)." [PSI:QC]
+comment: The protein coverage can provide insight into the sensitivity of the instrument/identification/protocols used.
+is_a: MS:4000005 ! table
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_column MS:1001093 ! sequence coverage
+relationship: has_column MS:1000885 ! protein accession
+relationship: has_column PEFF:0001006 or term above ???
+relationship: has_optional_column MS:1001197 ! DB composition target+decoy ???
+relationship: has_optional_column MS:1002153 ! protein level PSM counts
```
julianu: I assume the term "AA sequence length" is only relevant for this, so add it as a relationship?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-name: Pump pressure chromatogram
-def: "Representation of chromatographic pressure versus time. The first column contains the retention time points in seconds, the second column the corresponding pressures in pascal." [PSI:QC]
-is_a: MS:4000006 ! table
-is_a: MS:4000010 ! ID free
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:1003019 ! pressure chromatogram
-relationship: has_relation MS:4000013 ! single run based
-relationship: has_column: MS:4000108 ! Retention time
-relationship: has_column: MS:4000261 ! Pressure
+name: sampling rates
```
julianu: "peptide sampling rates"
Is this charge dependent, or leave this to the implementation? Are PTMs regarded or not (probably that's why you added the ??? to the respective relationships)?
I would rather leave it open to an implementation whether it is or not.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 [Term]
 id: MS:4000080
-name: Peptide elution order trailing
-def: "N(a, b) are the peptides common in the run and a reference run. N(a,b)_i to N(a,b)_j are the first to last eluting common peptides in the run ordered by retention time. This metric is the maximum of ('elution rank' of peptide N(a,b)_t in run) - ('elution_rank' of peptide N(a,b)_t in reference) for the first tenth of common peptides ordered by retention time. The metric is finally normalised with dividing by the number of (identified) peptides in the run." [PSI:QC]
-comment: The elution rank difference gives you a measure of how many peptides are eluting early.
-is_a: MS:4000003 ! single value
-is_a: MS:4000009 ! ID based
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:4000014 ! multiple runs based
+name: fraction of MS2 identified in different MS1 max quantiles
```
julianu: "after FDR" to the name?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+synonym: "MS2-4B" NARROW []
+synonym: "MS2-4C" NARROW []
+synonym: "MS2-4D" NARROW []
 
 [Term]
 id: MS:4000081
-name: Peptide elution order preceding
-def: "N(a, b) are the peptides common in the run and a reference run. N(a,b)_i to N(a,b)_j are the first to last eluting common peptides in the run ordered by retention time. This metric is the maximum of ('elution rank' of peptide N(a,b)_t in run) - ('elution_rank' of peptide N(a,b)_t in reference) for the last tenth of common peptides ordered by retention time. The metric is finally normalised with dividing by the number of (identified) peptides in the run." [PSI:QC]
-comment: The elution rank difference gives you a measure of how many peptides are eluting late.
+name: IQR RT period for identifications
```
julianu: Add "after FDR" or "after filtering" to the name?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-comment: A larger value indicates that isolation windows with that particular target m/z contributed more to the total TIC.
-xref: value-type:xsd\:double "The allowed value-type for this CV term."
-is_a: MS:4000004 ! n-tuple
-is_a: MS:4000010 ! ID free
-is_a: MS:4000024 ! MS2 metric
-relationship: has_relation MS:4000013 ! QC metric relation: one run
-relationship: has_relation MS:4000070 ! MS2 Total ion current chromatogram
-relationship: has_relation MS:4000082 ! isolation window target m/z
+name: mean precursor charge for MS2 spectra
+def: "Mean precursor charge for MS2 spectra" [PSI:QC]
```
julianu: "Mean precursor charge of all recorded MS2 spectra"

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-def: "For each unique isolation window this value reports the number of ions detected on average throughout the run." [PSI:QC]
-comment: A higher value indicates that a higher number of ions was detected for this specific isolation window on average.
-xref: value-type:xsd\:double "The allowed value-type for this CV term."
-is_a: MS:4000004 ! n-tuple
-is_a: MS:4000010 ! ID free
-is_a: MS:4000024 ! MS2 metric
-relationship: has_relation MS:4000013 ! QC metric relation: one run
-relationship: has_relation MS:4000082 ! isolation window target m/z
+name: median precursor charge for MS2 spectra
+def: "Median precursor charge for MS2 spectra" [PSI:QC]
```
julianu: "Median precursor charge of all recorded MS2 spectra"

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-comment: A higher value indicates greater variability in the number of ions detected for that particular isolation window.
-xref: value-type:xsd\:double "The allowed value-type for this CV term."
-is_a: MS:4000004 ! n-tuple
-is_a: MS:4000010 ! ID free
-is_a: MS:4000024 ! MS2 metric
-relationship: has_relation MS:4000013 ! QC metric relation: one run
-relationship: has_relation MS:4000082 ! isolation window target m/z
-relationship: has_relation MS:4000086 ! average isolation window density
+name: identified precursor median m/z
+def: "Median m/z value for all identified peptides (unique ions) at given FDR threshold." [PSI:QC]
```
julianu: Do you mean PSMs by "peptides (unique ions)"?
Add "after FDR" to the name.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-is_a: MS:4000006 ! table
-relationship: has_column MS:4000114 ! Sample rate column
-relationship: has_column MS:4000115 ! Count column
-relationship: has_optional_column MS:4000116 ! Peptide sequence column
-
-[Term]
-id: MS:4000121
-name: Explained precursor intensity first quarter
-def: "Fraction of identified MS2 in the first quarter of MS2 sorted by precursor intensity." [PSI:QC]
+name: explained precursor intensity
```
julianu: "fraction of explained precursor intensitiy quartiles"
Add "after FDR"

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+def: "Fraction of identified MS2 in the first, second, thrid, and forth quarter of MS2 sorted by precursor intensity at given FDR threshold." [PSI:QC]
```
julianu: typo in "third"

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-is_a: MS:4000009 ! ID based
-is_a: MS:4000001 ! QC metric
 synonym: "MS2-4D" NARROW []
 
 [Term]
-id: MS:4000125
-name: Extent of identified precursor intensity
-def: "Ratio of 95th over 5th percentile of precursor intensity for identified peptides" [PSI:QC]
+id: MS:4000089
+name: extent of explained precursor intensity
```
julianu: Add "after FDR"?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_metric_category MS:4000008 ! ID based metric
 relationship: has_units STATO:0000300 ! dimensionless ratio
 synonym: "MS1-3A" NARROW []
 
 [Term]
-id: MS:4000126
-name: Precursor intensity range of identified MS2
-def: "Minimum and maximum precursor intensity recorded and identified." [PSI:QC]
+id: MS:4000090
+name: explained precursor intensity range
```
julianu: Add "after FDR"?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 
 [Term]
-id: MS:4000150
-name: MS1 ion collection time low outliers (<Q1-1.5*IQR)
-def: "From the distribution of ion injection times (MS:1000927) for MS1, the list of outliers below Q1-1.5*IQR" [PSI:QC]
-comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
-is_a: MS:4000001 ! QC metric
-is_a: MS:4000010 ! ID free
+id: MS:4000097
+name: charged identifications ratio
```
julianu: "ratio of 1+/3+/4+ to 2+ charged identifications"
Are only these 3 recorded here, or "ratio of charged identifications to 2+ charged"?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 
 [Term]
-id: MS:4000151
-name: MS1 ion collection time high outliers (>Q3+1.5*IQR)
-def: "From the distribution of ion injection times (MS:1000927) for MS1, the list of outliers above Q3+1.5*IQR" [PSI:QC]
-comment: Injection time distribution can be used to gauge the suitability of used instrument settings for the sample content used.
-is_a: MS:4000001 ! QC metric
-is_a: MS:4000010 ! ID free
+id: MS:4000098
+name: charged spectra ratio
```
julianu: "ratio of 1+/3+/4+ to 2+ charged spectra"
Only these three charge states?

julianu (changes requested):
I added quite some comments.
At some point I stopped to add the "FDR filtered" to the name, maybe discuss this in the next call?

cv/qc-cv.obo
cbielow: I would add it should be `number of PSM**s**` ?!
and while unfiltered as a word is nice, this may lead to semantic issues, because sometimes you do have prefiltered results, e.g. when using XTandem's `maximum valid e-value` filter.
Also these sequences are prefiltered by (tryptic) specificity etc.. (but this is splitting hairs now :P)

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-comment: The metric enables to check visually e.g. MS2 TIC absolute value compared to MS1 total ion chromatogram or other analyses MS2 TICs.
-is_a: MS:4000006 ! table
-is_a: MS:4000010 ! ID free
-is_a: MS:4000022 ! chromatogram metric
-is_a: MS:4000024 ! MS2 metric
-relationship: has_relation MS:4000013 ! single run based
-relationship: has_relation MS:4000067 ! Total ion current chromatogram
-relationship: has_column: MS:4000108 ! Retention time
-relationship: has_column: MS:4000109 ! Relative intensity
+name: number of identified spectra after FDR
```
cbielow: should we mention MS level here?
usually this means MS2, but would an MS1 spectrum + mass fingerprinting also count as 'identified' ?

cbielow: How do we deal with ion mobility? does every 'microscan' in an IM frame count? What if scans were accumulated to improve S/N to obtain an ID?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_units UO:0000189 ! count unit
 
 [Term]
 id: MS:4000072
-name: Interquartile RT period for peptide identifications
-def: "The interquartile retention time period, in seconds, for all peptide identifications over the complete run." [PSI:QC]
-comment: Longer times indicate better chromatographic separation.
+name: number of identified distinct peptides
+def: "Number of identified peptides without duplicate count (or modifications) at given FDR threshold." [PSI:QC]
```
cbielow: overlapping peptides are counted separately, I assume, e.g. a missed cleavage is a new peptide?!

cv/qc-cv.obo
cbielow: we don't mind how grouping is done, right? Maybe mention that...

cv/qc-cv.obo
cbielow: not sure what this means...
parsimony?

cv/qc-cv.obo
cbielow: also not sure here. Expand definition.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-name: Median MS1 peak FWHM for peptides in RT quantile
-def: "Median of all MS1 peak widths at half maximum (FWHM) for all identified peptide, in seconds, in the respective identifications' quantiles ordered by retention time. The data is given in a tuple defining the quantiles, 4-tuple represents quartiles, 10-tuple deciles, etc." [PSI:QC]
-comment: Sharper peak widths (i.e. smaller values) indicate better chromatographic resolution in the given quantile.
-is_a: MS:4000004 ! n-tuple
-is_a: MS:4000009 ! ID based
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:1000086 ! full width at half-maximum
-relationship: has_relation MS:4000013 ! single run based
+name: ID ratio
+def: "The ratio of identified and recorded MS2 spectra at given FDR threshold." [PSI:QC]
```
cbielow: value between 0-1?
(people could mistake this to be 0-100?)

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_units STATO:0000300 ! dimensionless ratio
 
 [Term]
 id: MS:4000077
-name: Area under TIC
-def: "The area under the total ion chromatogram." [PSI:QC]
+name: AA sequence length
+def: "The length of an amino acid sequence." [PSI:QC]
```
cbielow: The definition does not add value.
The **number** of amino acids in the sequence?

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
+name: protein coverage table
+def: "The coverage of protein sequences from the peptide sequences identified. The table records the coverage itself (in percent), the protein accession (SHOULD correspond to the accession used in fasta document used for Sequence DB based identification), length of the protein (, and optionally if it is 'target' or 'decoy' entry and how many PSM can be attributed to this protein)." [PSI:QC]
+comment: The protein coverage can provide insight into the sensitivity of the instrument/identification/protocols used.
+is_a: MS:4000005 ! table
+relationship: has_metric_category MS:4000008 ! ID based metric
+relationship: has_relation QC:4000012 ! single run based metric
+relationship: has_column MS:1001093 ! sequence coverage
+relationship: has_column MS:1000885 ! protein accession
+relationship: has_column PEFF:0001006 or term above ???
+relationship: has_optional_column MS:1001197 ! DB composition target+decoy ???
```
cbielow: i think this term only defines the state of the FASTA DB as a whole (e.g. its a combined db).
It does not define the allowed values, `target` and `decoy` AFAICT.

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
-def: "Representation of chromatographic pressure versus time. The first column contains the retention time points in seconds, the second column the corresponding pressures in pascal." [PSI:QC]
-is_a: MS:4000006 ! table
-is_a: MS:4000010 ! ID free
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:1003019 ! pressure chromatogram
-relationship: has_relation MS:4000013 ! single run based
-relationship: has_column: MS:4000108 ! Retention time
-relationship: has_column: MS:4000261 ! Pressure
+name: sampling rates
+def: "The sampling rates of identified MS2 and their respective frequency. The table records the sampling rate (i.e. once=1, twice=2, ...) in one column and the respective frequency of that sampling rate in another column. Optionally, another column giving the sequences of peptides sampled (space separated) can be present." [PSI:QC]
```
cbielow: the name should be more specific
and I'm not sure what the metric means and how its computed :)
My guess: one row per peptide sequence, where the sampling rate says how many times an eluting peptide was triggered for MS2 (i.e. potential oversampling).
But would you include unidentified MS2 spectra from the same mass trace? (probably not, also a bit harder to compute)

how do you compute the frequency? What is the denominator?

While this metric is nice, it will be a BIG table.
How about a separate metric which just holds some aggregration stats?
See https://htmlpreview.github.io/?https://github.com/cbielow/PTXQC/blob/master/inst/examples/report_v1.0.10__txt_5files_withMatch-100min.html#evd-oversampling-msms-counts-per-3d-peak

cv/qc-cv.obo
```diff
@@ -18,1967 +18,341 @@ import: https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/1.4/stato
...
 id: MS:4000080
-name: Peptide elution order trailing
-def: "N(a, b) are the peptides common in the run and a reference run. N(a,b)_i to N(a,b)_j are the first to last eluting common peptides in the run ordered by retention time. This metric is the maximum of ('elution rank' of peptide N(a,b)_t in run) - ('elution_rank' of peptide N(a,b)_t in reference) for the first tenth of common peptides ordered by retention time. The metric is finally normalised with dividing by the number of (identified) peptides in the run." [PSI:QC]
-comment: The elution rank difference gives you a measure of how many peptides are eluting early.
-is_a: MS:4000003 ! single value
-is_a: MS:4000009 ! ID based
-is_a: MS:4000022 ! chromatogram metric
-relationship: has_relation MS:4000014 ! multiple runs based
+name: fraction of MS2 identified in different MS1 max quantiles
+def: "The fraction of all MS2 scans identified at given FDR threshold in the respective quantile of the sorted list of parent MS1 maximum intensity." [PSI:QC]
```
cbielow: how do you define the FDR threshold (also see similar metrics below)

cv/qc-cv.obo
cbielow: is that the precursor intensity as annotated by the instrument (e.g. the preceding MS1 scan), or an abundance computed by a software for the whole elution&isotope profile of an ion species?

bittremieux:

PR #170, opened 2022-02-15
packaged ID-based metrics 

mwalzer:
(mostly distribution descriptions) for another PR into MS-CV.
Distributions described by Q1,2,3, mean, sigma, 1.5*IQR outliers - all separate as we discussed some reports can not/want not to report all values.
The proper accessions need to be replaced as soon as this PR is up in the queue.


bittremieux:

PR #202, opened 2022-07-05
Website changes

mwalzer:
Website changes as discussed in the last telcos.
This brings changes in the repo, too. Most importantly:
* gh-page directly in main and directly in docs folder
* new dedicated folder for specification docs
* examples moved into specification docs folder (to keep permalinks correct through versions)
* worked examples, companion documents, etc. are now living in docs/pages
* gh-page **_cannot_** pull anything outside of the docs folder into the website so spec doc (and maybe in the future others) need to be updated by way of copy to docs/pages/from_repo


mwalzer:
All the above made keeping correct permalinks easier to keep but needed change in the spec doc. The spec doc needed link updates in any case - but this is why I am fast-tracking this PR. That version of the spec doc is going to be the active draft in the current review round.

PR #219, opened 2024-03-19
Update set mzQC example

bittremieux:
To update the mzQC set example, we need a few additional changes:
- There is no term for the old [protein contaminant intensity ratio](https://github.com/HUPO-PSI/mzQC/blob/main/cv/qc-cv.obo#L1941). I added an [updated definition in the mzQC file](https://github.com/HUPO-PSI/mzQC/compare/intro_set?expand=1#diff-4ccd55deb8e19f9857e799230f8eaff6e654bc8895f3089850d9b22d8823b27eR78) itself, but I'm not familiar with this metric, so it should be double-checked. **TODO:** This still needs to be requested in the CV and then updated with the corresponding accession number here.
- I updated the PCA table. This seems fine, except that it's an ID-based metric, so an ID input file has to be added. I've included the `proteinGroups.txt` based on my limited knowledge of MaxQuant and the previous description. Does this make sense? There doesn't seem to be a dedicated entry for MaxQuant output files in the CV yet, so I've used a more generic one.
    - The validator currently doesn't recognize this as an ID file.


docs/pages/worked-examples/intro_set.md
```diff
@@ -0,0 +1,168 @@
...
+    ...
+  }
+],
+```
+
+Let's dive into an example metric, like the "protein contaminant intensity ratio," indicating how much of your sample is taken up by known contaminants. A higher value suggests more contamination:
+
+```
+{
+  "accession": "MS:4000XXX",
```
cbielow, line 116: lets wait for https://github.com/HUPO-PSI/psi-ms-CV/pull/246 to get a definitive number here

specification_documents/examples/intro_set.mzQC
```diff
@@ -308,15 +329,10 @@
...
-      {
-        "name": "Proteomics Standards Initiative Quality Control Ontology",
-        "uri": "https://github.com/HUPO-PSI/mzQC/blob/main/cv/qc-cv.obo",
-        "version": "1.0.0"
-      },
       {
         "name": "Proteomics Standards Initiative Mass Spectrometry Ontology",
-        "uri": "https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v4.1.71/psi-ms.obo",
-        "version": "4.1.71"
+        "uri": "https://github.com/HUPO-PSI/psi-ms-CV/releases/download/v4.1.130/psi-ms.obo",
```
cbielow, line 334: needs whatever was used in https://github.com/HUPO-PSI/psi-ms-CV/pull/246 or later

bittremieux:
I updated the example. @cbielow could you give it another review before merging?

One thing to discuss is that I added the fraction unit for the protein contaminant fraction metric. However, in the CV [this term](https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L24122) doesn't have a unit specified. So I think we should still add it there.

(Side note: the validator doesn't complain about this. Is this something that should be checked? @mwalzer)

Another issue is that the file doesn't validate fully because it contains ID-based metrics, but the validator complains that we haven't listed any ID files as input.
1. For the "all" set, we have included the proteinGroups.txt file, but the validator doesn't recognize this. @mwalzer can this be updated?
2. For the "healthy" and "diseased" sets, we indeed don't have an ID file listed. Should this be the same proteinGroups.txt file from MaxQuant @cbielow?

docs/pages/worked-examples/intro_set.md
```diff
@@ -0,0 +1,214 @@
...
+mzQC can store PCA results, capturing the variation between these two states.
+
+For this we extracted protein abundances from the `proteinGroups.txt` file specified as an input file to the "all" set.
+This file is produced by MaxQuant and contains quantified protein intensities along with other identification information for each protein group detected in the experiment.
+
+First, let's have a look at what the PCA plot would look like, plotting the first two principal components:
+
+![PCA plot of the healthy vs diseased samples.](../../pages/figures/intro_set_pca.png)
+
+Next, we'll look at how mzQC can encapsulate such analysis, storing the the first five principal components as a table metric, referenced by the previously defined set labels.
```
cbielow, line 170:
```suggestion
Next, we'll look at how mzQC can encapsulate such analysis, storing the first five principal components as a table metric, referenced by the previously defined set labels.
```

cbielow (approved):

cbielow:
> One thing to discuss is that I added the fraction unit for the protein contaminant fraction metric. However, in the CV [this term](https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L24122) doesn't have a unit specified. So I think we should still add it there.

Agreed.

> 2.    Is this the correct input file considering that the metric we're using is "principal component analysis of MaxQuant's protein group raw intensities"? @cbielow I'm not familiar enough with MaxQuant's outputs. We also have another CV term that describes PCA results of MaxQuant's group lfq intensities.

Yes, proteinGroups.txt is correct. Which CV term you pick is arbitrary here, since the numbers are purely for illustration. Can stay as is IMHO.


> For the "healthy" and "diseased" sets, we indeed don't have an ID file listed. Should this be the same proteinGroups.txt file from MaxQuant @cbielow?

Yes.

bittremieux:
Ready to be merged after [psi-ms-CV#303](https://github.com/HUPO-PSI/psi-ms-CV/pull/303) is merged to ensure that the CV version is accurate.

PR #228, opened 2024-07-09
update QC metrics in mzML files example

bittremieux:
There's still one small issue with this example. The [spectra half-TIC metric](https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo#L22927) that we're using is defined for MS2 spectra in the CV, but in this mzML file the three spectra are actually MS1 spectra. So semantically this is invalid, although no parser will flag this issue.

Should we?

- Update the mzML file to contain MS2 spectra instead. This will be confusing imo, because to be valid, there also needs to be an MS1 spectrum first and then we need to explain this tricky difference.
- Remove the "MS2 metric" relationship and have this term instead be applicable to any level of MS. This works reasonably I think, the only issue could be that the metric is not really informative for profile spectra (which are typically MS1).


bittremieux:
CV updated requested in [psi-ms-CV#304](https://github.com/HUPO-PSI/psi-ms-CV/pull/304).

This PR should be merged after #219.

PR #229, opened 2024-07-10
updated USI example to mzQC v1.0

cbielow:
- added a few more references (for USI, and CV terms)
- fixed links (ProteomicsDB keeps failing -- its their API according to their support team; PRIDE works ATM)
- arranged columns in the CV Table as columns (would show as one row, which might confuse people)

The validator is still not entirely happy, but I think it needs an update of the CV:
```
    Metric use of severity 5 and message: Non-metric CV term used in metric context: accession = MS:4000068
    Metric value no-unit of severity 3 and message: Metric CV term used without value unit specification. accession(s) = MS:4000068

    Loading online vocabulary of severity 5 and message: Error loading the following online ontology referenced in mzQC file: could not find a parser to parse 'https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo'
    Loading online vocabulary of severity 5 and message: Error loading the following online ontology referenced in mzQC file: HTTP Error 404: Not Found

    Unknown CVTerm of severity 7 and message: CV term used without matching ontology entry: accession = MS:1001058
    Unknown CVTerm of severity 7 and message: CV term used without matching ontology entry: accession = MS:4000068
```


mwalzer:
mzQC should be good now, validator is not complaining anymore.
I used a new version of psi-ms and a 'stable' repo link to a release version of the UO github tree (unless they revise their git history, that link should work in the future)

bittremieux (approved):

PR #236, opened 2024-07-24
Remove QC-CV

bittremieux:
It was never part of an official standard, so the legacy QC-CV can be removed to avoid confusion for users.


mwalzer (approved):
Lets get rid of the old distractions

PR #242, opened 2024-08-21
Linkcheck on PR

cbielow:
fixes  https://github.com/HUPO-PSI/mzQC/issues/241


cbielow:
seems to work just fine :)

PR #323, opened 2025-11-05
Documentation on new metrics classification

bittremieux:
Documentation related to the formalized metrics classification [proposed during the ELIXIR BioHackathon](https://github.com/MS-Quality-Hub/biohackathon2025/issues/4) and [discussed in the PSI-MS CV here](https://github.com/HUPO-PSI/psi-ms-CV/issues/430).

- Add new documentation page with details on the metrics classification.
- Update CV usage and term request guide to reflect this.

### DIFF
--- a/docs/pages/cv/classification_reference.md
+++ b/docs/pages/cv/classification_reference.md
@@ -126,6 +126,8 @@ Specifies which type of data input the metric requires to be computed.
 
 * **Raw acquisition data:** metrics that can be calculated directly from the raw MS data, without identifications.
   *Example:* total ion current stability, scan count.
+* **Deconvoluted data:** metrics based on processed spectra or peak lists obtained after signal deconvolution, centroiding, or deisotoping, but prior to identification.
+  *Example:* peak density in deconvoluted spectra, precursor mass range coverage.
 * **Identification results:** metrics that depend on identified peptides, compounds, or spectra.
   *Example:* PSM-level FDR, peptide coverage.
 * **Quantification results:** metrics derived from quantitative data matrices.
--- a/docs/pages/cv/classification_reference.md
+++ b/docs/pages/cv/classification_reference.md
@@ -1,0 +1,223 @@
+---
+layout: page
+title: "Metrics – Classification Reference"
+permalink: /metrics/classification
+---
+
+# QC Metric Classification Reference
+
+*Standardized semantic categories for PSI-MS quality control metrics*
+
+## Overview
+
+Each QC metric in the PSI-MS Controlled Vocabulary (CV) is annotated using **seven independent classification dimensions**.
+Together, these describe *what a metric measures*, *where it applies in the workflow*, and *how it should be interpreted*.
+
+Every metric must define **exactly one relationship** from each dimension.
+This ensures complete, machine-interpretable semantics across all QC terms used in mzQC and related standards.
+
+| Dimension                       | Describes                                                              |
+| ------------------------------- | ---------------------------------------------------------------------- |
+| **Workflow stage**              | Where in the experimental or computational pipeline the metric applies |
+| **Analytical dimension**        | What fundamental property the metric measures                          |
+| **Information dependency type** | What type of data the metric depends on                                |
+| **Measurement scope**           | At what aggregation level the metric summarizes data                   |
+| **Acquisition strategy**        | Which acquisition or instrument configuration it applies to            |
+| **Quality interpretation type** | How metric values relate to data quality                               |
+| **Metric value type**           | How metric values are structurally represented (single, tuple, etc.)   |
+
+## Workflow Stage
+
+**Definition:**
+The experimental or computational stage of the workflow to which a QC metric applies.
+
+This tells *where* in the process the metric is relevant — from sample preparation through acquisition and analysis.
+
+### Subclasses
+
+#### Experimental workflow stage
+
+Metrics describing quality at the laboratory or instrument level.
+
+* **Sample preparation stage:** metrics describing sample handling, labeling, digestion, or storage quality.
+  *Example:* peptide recovery yield.
+* **Chromatography stage:** metrics about LC separation performance.
+  *Example:* retention-time reproducibility, peak width.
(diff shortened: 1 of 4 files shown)


mwalzer:
I think it would also be good to document from which CV version this is consistently implemented?

docs/pages/cv/classification_reference.md
```diff
@@ -0,0 +1,297 @@
...
+
+* **Acquisition mode independent:** metrics valid for any acquisition method.
+* **Data-dependent acquisition (DDA):** metrics specific to stochastic precursor selection workflows.
+  *Example:* number of MS2 spectra per precursor.
+* **Data-independent acquisition (DIA):** metrics for window-based fragmentation strategies.
+  *Example:* precursor window purity.
+* **Targeted acquisition:** metrics for SRM, PRM, or other targeted workflows.
+  *Example:* transition reproducibility.
+* **Ion-mobility-coupled metric:** metrics derived from acquisition methods that include gas-phase ion mobility separation.
+  *Example:* TIMS mobility resolution (Δ1/K₀) per run.
```
cbielow: not sure if an explicit `Ion-mobility` value is needed/helpful, because it is orthogonal to DIA vs DDA,
and to be complete, we'd also need `LC coupled metric` (since this also adds another dimension, just as IM does) vs. direct injection.
We are mixing concepts here IMHO.
Aquisition mode (referring to the MassSpec itself) may just be
1) any
2) DDA
3) DIA
4) Targeted

other things like: LC, IM, Imaging are a different thing??
Even the specialized MSn is tricky, since TMT can be on MS3 level, yet still is DDA.

Now, one could just add multiple values for this dimension, but maybe we should push concepts like LC, IM etc into `workflow stage`? (where it would be duplicated anyways?)

docs/pages/cv/classification_reference.md
```diff
@@ -0,0 +1,297 @@
...
+- **Ionization quality metric:** properties of the precursor ion population (e.g., charge-state distribution, adduct prevalence).
+- **Ion mobility metric:** IMS resolution, drift-time/CCS accuracy and reproducibility.
+- **Spectral quality metric:** quality of individual spectra (e.g., peak density, S/N, completeness).
+- **Fragmentation efficiency metric:** effectiveness of precursor ion fragmentation to produce interpretable spectra.
+- **Isolation purity metric:** precursor isolation selectivity or co-isolation of interfering species.
+- **Identification confidence metric:** reliability of identifications (e.g., FDR, ID rate).
+- **Quantification precision metric:** reproducibility or variability of quantitative results.
+- **Contamination metric:** unwanted signal from contaminants, carryover, or background.
+- **Instrument operational performance metric:** general indicators of instrument health (e.g., vacuum, detector voltage, temperature).
+- **Missingness/completeness metric:** data absence or completeness across features, runs, or studies.
```
cbielow: this has quite considerable overlap with the `workflow` stage and it's potentially a dimension which is hard to cover completely, since to me it's quite fuzzy. This may lead to people cramming their metric into something which "fit's good enough" instead of thinking of a new category value.

What could be a set of values which is
1) easy to categorize
2) does not need extension in the future
3) adds value compared to the other dimensions

, or in other words: what could this dimension represent, making it distinct from others?
(so dimensions are ideally orthogonal)

docs/pages/cv/classification_reference.md
```diff
@@ -0,0 +1,297 @@
...
+#### Subclasses
+
+**Experimental workflow stage**
+
+Metrics describing quality at the laboratory or instrument level.
+
+* **Sample preparation stage:** metrics describing sample handling, labeling, digestion, or storage quality.
+  *Example:* peptide recovery yield.
+* **Chromatography stage:** metrics about LC separation performance.
+  *Example:* retention-time reproducibility, peak width.
```
cbielow: one could add `Direct injection stage`

docs/pages/cv/classification_reference.md
```diff
@@ -0,0 +1,297 @@
...
+* **Higher is better:** increasing values indicate improved quality.
+  *Example:* identification rate, mass accuracy score.
+* **Lower is better:** decreasing values indicate improved quality.
+  *Example:* FDR, mass error.
+* **Context dependent:** interpretation varies depending on method or range.
+  *Example:* precursor charge-state fractions, peak density.
+* **Target range:** optimal quality corresponds to values within a defined interval.
+  *Example:* temperature, pressure, retention-time drift.
+* **Categorical:** quality expressed as discrete categories (e.g., pass/fail, OK/warning/error).
+* **Trend:** metrics intended for temporal monitoring rather than direct ranking (e.g., instrument drift over time).
```
cbielow: I find `trend` would fit into one of the other categories, e.g. `instrument drift over time` fits into `lower is better`.
Are there any counter examples where this is not the case?

cbielow (commented):
this came out more tricky than anticipated.

More input is highly appreciated!

docs/pages/cv/classification_reference.md
```diff
@@ -0,0 +1,297 @@
...
+
+**Definition:**
+The experimental or computational stage of the workflow to which a QC metric applies.
+
+This tells *where* in the process the metric is relevant — from sample preparation through acquisition and analysis.
+
+#### Subclasses
+
+**Experimental workflow stage**
+
```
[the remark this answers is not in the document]
bittremieux: Data analysis, as you only have the number of proteins after data analysis?

docs/pages/cv/classification_reference.md
bittremieux: I agree that this mixes several orthogonal concepts: precursor selection logic (DDA/DIA/targeted), physical separation (LC, IM), and acquisition strategies (imaging, MSn)

Maybe a more coherent approach is to redefine this dimension strictly as "how ions are selected for fragmentation or measurement." In other words, capturing only the precursor selection logic of the mass spectrometer, independent of chromatography, IM, imaging, etc.

In that case, we can simplify the subclasses to:
- Acquisition-mode independent
- Data-dependent acquisition (DDA)
- Data-independent acquisition (DIA)
- Targeted acquisition

